Issue #182, opened 2026-06-02
Reduced-precision (bf16/fp16) decode produces garbage in C++ runtime for vision_language/embed_input bundles


samli-gpu:
## Summary

The native C++ runtime (`trtmc run`) produces **garbage output at `--precision bf16` and `fp16`** for **vision-language / `embed_input` decoder bundles** (the non-split, single decoder-engine path). The **TRT engine itself is correct** at those precisions — the same bundle runs coherently through the pure-Python runner (`tensorrt_model_connect.debug_runner`) — so this is a **runtime execution bug**, not a builder/engine numerics problem. `--precision fp32` works correctly end-to-end.

This is **not model-specific**. It reproduces on stock `Qwen/Qwen2.5-VL-3B-Instruct`. (Found while adding a new `vision_language` family; that family is pinned to fp32 as a workaround.)

## Reproduce

```bash
# bf16: garbage
./build/trtmc build Qwen/Qwen2.5-VL-3B-Instruct -o /tmp/q.trtfb --max-cache-length 384 --precision bf16
./build/trtmc run /tmp/q.trtfb --prompt "What color is the vehicle? Answer in one word." \
  --image tests/e2e/data/test_img.jpeg --max-new-tokens 12 --greedy
# -> "This weekend problem solving for future information below the problem>SOC"

# fp32: correct
./build/trtmc build Qwen/Qwen2.5-VL-3B-Instruct -o /tmp/q32.trtfb --max-cache-length 384 --precision fp32
./build/trtmc run /tmp/q32.trtfb --prompt "What color is the vehicle? Answer in one word." \
  --image tests/e2e/data/test_img.jpeg --max-new-tokens 12 --greedy
# -> correct, concise answer
```

The same **bf16** `engine_plan` from the bundle runs **coherently** via Python:

```python
from tensorrt_model_connect.debug_runner import runner_from_bundle
r = runner_from_bundle("/tmp/q.trtfb")
r.generate(input_ids, 12)   # coherent text
```

## What is ruled out

Verified during investigation that the following are **not** the cause:

- **Tokenizer** — the C++ native BPE emits the same ids as HuggingFace.
- **Logits dtype** — logits are cast to fp32 in the builder; the runtime reads fp32 correctly.
- **KV-cache write** — byte-verified that the per-step present→cache copy is correct (`present_bytesum == cache_bytesum` at every step; element size = 2 for bf16).
[...]

PR #183, opened 2026-06-02
feat(lance): add Lance image/video understanding (x2t)

samli-gpu:
## Background

TRT-MC had no support for [`bytedance-research/Lance`](https://huggingface.co/bytedance-research/Lance), a 3B "native unified multimodal" model (image+video understanding, generation, and editing) built on Qwen2.5-VL-3B with a Qwen2.5-VL ViT and a Wan2.2 VAE. This PR adds the **understanding** path (image/video → text), which is the most tractable capability and maps cleanly onto the existing `vision_language` runtime strategy.

Rebased on top of #184 (`fix(vl): honor reduced precision for embed bundles`), which this PR depends on for bf16 embed-bundle correctness.

## Exit Criteria

- A `lance` family plugin builds a `.trtfb` bundle and the `vision_language` runtime answers image-understanding questions correctly at bf16.
- Numerical correctness validated against an independent reference.
- `scripts/check_family_coverage.py` stays green.
- Non-goal: generation/editing (`t2i`/`t2v`/`image_edit`/`video_edit`) is explicitly out of scope here (it needs a new runtime strategy).

## Implementation

- **`python/tensorrt_model_connect/families/lance/`** — the family plugin (`name="lance"`, `runtime_strategy="vision_language"`, `embed_input=True`). Lance is a Mixture-of-Transformer-Experts checkpoint; the loader reads only the **understanding-expert** weights via `load_standard_weights(model_prefix="language_model.model", ...)`, which naturally drops the `*_moe_gen` generation expert plus the `llm2vae`/`vae2llm`/`time_embedder`/`latent_pos_embed` tensors. It reuses the Qwen2.5-VL decoder builder (QKV-bias + per-head QK-norm, applied conditionally) and the Qwen2.5-VL ViT builder (re-adding the `visual.` prefix Lance ships bare). `build_engine` forwards `precision` so reduced-precision embed bundles build correctly on top of #184.
- **`scripts/prepare_lance_model.py`** — the Lance HF repo is not a flat checkpoint (it nests `Lance_3B/llm_config.json` and a separate `Qwen2.5-VL-ViT/` dir). This stages a buildable directory: `config.json` (with `model_type` stamped `lance` so it routes to this plugin instead of `qwen_vl`), the LLM `model.safetensors`, tokenizer files, and the ViT at `vision/model.safetensors`.
- **`tests/e2e/models/lance-3b-x2t-image.json`** — E2E manifest for `x2t_image` (satisfies the coverage gate), `skip_reason`-gated for now.
- **`tools/test_impact_fallback_allowlist.txt`** — registers the staging helper as `no_impact` (one-off script, not imported by package/tests) so the CI impact-analysis gate passes.

## Validation

- `ruff check python/tensorrt_model_connect/families/lance/ scripts/prepare_lance_model.py`: **passed**.
- `python3 scripts/check_family_coverage.py`: **green** — `[OK] lance (1 model)`.
- `python3 tools/test_impact.py --validate`: **passed** (196 models, 12 core, 73 families).
- End-to-end `./build/trtmc run` at **bf16** on `tests/e2e/data/test_img.jpeg`:
  - "Describe this image in one sentence." → **"White car driving on the street."**
  - "What color is the vehicle? Answer in one word." → **"White"**
  - (bf16 decoder engine is ~6.5 GB vs ~13 GB fp32, confirming true reduced precision.)
- TRT decoder vs an **independent eager PyTorch reference**: **exact match** per-layer and per-logit.

## Notes For Future Readers

- **Why `skip_reason`:** the E2E harness builds `trtmc build <hf_id>` directly, but Lance needs `scripts/prepare_lance_model.py` staging first; and there is no HF AutoModel reference for Lance (`library_name=Lance`, custom `modeling.lance`), so a Lance reference backend is needed. Remove `skip_reason` once both are wired into the VL runner.
- **bf16** works after #184 (input_embed bound as fp32 and cast inside the graph) + forwarding `precision` here. The original reduced-precision bug was filed as #182.
- **Generation/editing** is a larger future stage: an LLM-as-flow-matching-denoiser runtime strategy + Wan VAE decode, reusing `families/wan_t2v` and the C++ flow-match scheduler.
- **Review order:** `families/lance/plugin.py` (loader + build reuse) → `scripts/prepare_lance_model.py` (staging contract) → the manifest.

### DIFF
--- a/python/tensorrt_model_connect/families/lance/__init__.py
+++ b/python/tensorrt_model_connect/families/lance/__init__.py
@@ -1,0 +1,6 @@
+"""Lance family package. Exposes the module-level ``plugin`` for auto-discovery."""
+from __future__ import annotations
+
+from .plugin import plugin
+
+__all__ = ["plugin"]
--- a/python/tensorrt_model_connect/families/lance/plugin.py
+++ b/python/tensorrt_model_connect/families/lance/plugin.py
@@ -1,0 +1,156 @@
+"""Lance family plugin — ByteDance ``bytedance-research/Lance`` unified model.
+
+Scope (Stage 1): the **understanding** path only — ``x2t_image`` and
+``x2t_video``. Lance's understanding sub-model is a Qwen2.5-VL ViT vision
+encoder feeding a Lance text decoder, which maps onto the existing
+``vision_language`` runtime strategy.
+
+Lance is a Mixture-of-Transformer-Experts model: every decoder layer carries a
+second ``*_moe_gen`` parameter set, plus ``llm2vae`` / ``vae2llm`` /
+``time_embedder`` / ``latent_pos_embed`` tensors. Those drive flow-matching
+image/video **generation** and are intentionally NOT consumed here:
+``load_standard_weights`` only reads the unsuffixed understanding-expert keys
+(``self_attn.q_proj``, ``mlp.*``, ``input_layernorm`` …), so the generation
+expert is dropped automatically. Generation/editing is a later stage that needs
+a new runtime strategy and is out of scope for this plugin.
+
+Architecture (confirmed against ``modeling/lance/qwen2_navit.py``): the
+understanding decoder is GQA (16/2) with **QKV bias** (Qwen2 style) **and**
+per-head **QK-norm** over ``head_dim`` (``qk_norm_und``) + SwiGLU + standard
+RoPE; ViT is the standard Qwen2.5-VL encoder shipped with bare ``blocks.*`` /
+``merger.*`` / ``patch_embed.*`` names (we re-add the ``visual.`` prefix the
+shared vision builder expects). The shared decoder builder applies QKV-bias and
+QK-norm conditionally when the weights are present.
+
+Numerical validation: the TRT decoder matches an independent eager reference
+exactly (per-layer and logits), and end-to-end ``trtmc run`` at **bf16** is
+verified correct ("White car driving on the street." / "White"). Reduced
+precision relies on the #184 builder fix (now in main): for embed bundles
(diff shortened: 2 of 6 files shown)


samli-gpu:
### CI status — builder-tier timeout (not a code defect)

All Premerge CI gates pass except one, which **times out** rather than fails on a test:

- ✅ Build pip package, Impact analysis, Build C++ tests, Check family coverage, Cyclomatic complexity, **Lint**, **C++ unit tests**
- ⏱️ **Python builder and tools tests** — *timed out after 40 min* on two consecutive runs (reaching ~93%). Every test running at the cutoff is a pre-existing builder test unrelated to this PR (segformer, mixtral, elf, standard_decoder, …).

Why this looks like a CI-capacity issue rather than this change:

- This PR adds **no** Python builder tests. Impact analysis selected the full `builder` unit-test tier (because a new family package landed) with `"builder_tests": []`; the Lance plugin is not exercised by any builder test (no `lance` test file, no real-registry enumeration that builds engines).
- A sibling new-family PR (`add-locateanything-support`) is hitting the same long-running/timeout behavior, while unrelated PRs pass — consistent with the full `builder` tier sitting right at the 40-min job cap for new-family selections on the current runners.

I've intentionally **not** modified any test, timeout, or selection rule to get past this. Could a maintainer re-run the job when runners are less contended (or advise if the `builder` tier needs a higher timeout / finer sharding for new-family PRs)?

Note: the `fp32` pin here is a workaround for the general reduced-precision runtime bug in #182, whose fix is already in flight in #184 (*fix(vl): honor reduced precision for embed bundles*). Once #184 lands, a follow-up can lift the pin and switch this manifest to bf16.

PR #196, opened 2026-06-03
feat(lance): native repo build + CI E2E coverage for x2t

samli-gpu:
## Background

#183 added Lance x2t image-understanding support, but its E2E manifest was `skip_reason`-gated, so CI **did not actually build, run, or validate** Lance — a coverage gap. It was skipped for two real reasons: the harness builds `trtmc build <hf_id>` directly while the Lance HF repo is not a flat checkpoint (it nests `Lance_3B/` + a separate `Qwen2.5-VL-ViT/`), and there is no HF AutoModel reference for Lance (`library_name=Lance`, custom `modeling.lance`). This PR closes both.

## Exit Criteria

- `tests/test_e2e.py::test_e2e[lance-3b-x2t-image]` **runs** (no longer skips): builds the bundle, runs the C++ binary on the test image, and validates the output.
- The check **gates on content** (fails on garbage), not just on non-empty output.
- No behavior change for non-Lance models.

## Implementation

- **Native repo build** (`engine_builder._resolve_model` + new `families/lance/staging.py`): `trtmc build bytedance-research/Lance` (or a local unpacked Lance repo dir) now stages the nested repo internally — `config.json` (model_type=lance), the LLM `model.safetensors`, tokenizer files, and the ViT at `vision/` — analogous to the existing `.nemo` handling. Cache-aware (resolves offline when cached) and a cheap no-op for non-Lance models. `scripts/prepare_lance_model.py` becomes a thin wrapper over the shared staging.
- **Reference oracle** (`tests/e2e/models/lance-3b-x2t-image.json`): drop `skip_reason`; use a `golden_snapshot` reference with a single `full_generation` (text) stage. The default `vision_language_generation` thresholds — `non_empty AND (normalized_text_edit_distance <= 0.5 OR token_agreement_rate >= 0.3)` — accept the correct bf16 answer and reject the reduced-precision garbage that motivated #182.
- **Golden** (`tests/e2e/golden/lance-3b-x2t-image/full_generation.json`): the validated answer text.

## Validation

- `pytest tests/test_e2e.py -k lance`: **PASSED** (158s) — built Lance via the harness, ran the C++ binary on `tests/e2e/data/test_img.jpeg`, compared to the golden.
- **Negative check**: flipping the golden to a wrong answer makes the same test **FAIL** (18s, reusing the cached bundle) — proving it gates on content, not just non-empty.
- `trtmc build bytedance-research/Lance` (remote id) and a local Lance repo dir both build + run → "White".
- `ruff`, `scripts/check_family_coverage.py`, `tools/test_impact.py --validate`, and `tests/builder/test_families.py` all pass.

## Notes For Future Readers

- The golden is a **regression snapshot**; the thresholds are case/whitespace tolerant (word-level `token_agreement_rate` is lowercased; NED is the alternative), so it's robust across runners while still catching gross output regressions.
- Native build is now the primary path; `prepare_lance_model.py` remains as an optional convenience.
- Generation/editing (t2i/t2v) is still out of scope.
- **Review order:** `families/lance/staging.py` → the two `engine_builder._resolve_model` hooks → the manifest + golden.

### DIFF
--- a/python/tensorrt_model_connect/engine_builder.py
+++ b/python/tensorrt_model_connect/engine_builder.py
@@ -357,6 +357,15 @@ def _resolve_model(model_id_or_path: str) -> str:
         if nemo_files:
             return _resolve_nemo_archive(nemo_files[0])
 
+    # Handle an unpacked Lance repo (nested Lance_3B/ + Qwen2.5-VL-ViT/, not a
+    # flat HF checkpoint) by staging it into a buildable dir. Cheap no-op for
+    # any non-Lance local dir.
+    if local.is_dir():
+        from .families.lance.staging import resolve_and_stage_lance
+        staged = resolve_and_stage_lance(str(local))
+        if staged is not None:
+            return staged
+
     # Treat as HuggingFace repo ID — download to HF cache.
     try:
         from huggingface_hub import snapshot_download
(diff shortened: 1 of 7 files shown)


samli-gpu:
Superseded by https://github.com/NVIDIA/TensorRT-Model-Connect/pull/215 — recreated from a branch in NVIDIA/TensorRT-Model-Connect so premerge CI can run (the CI router only accepts same-repo branches, not forks). Same commits and diff.

PR #210, opened 2026-06-05
fix(tests): skip torch-trt time-series e2e when trtmc binary is absent

### LINKED ISSUE #209: CI: torch-trt time-series e2e hard-asserts source-tree build/trtmc → false failures under broad test selection
## Summary

`tests/engine_defs/torch_trt/test_time_series_e2e.py::test_time_series_models_match_reference_end_to_end`
**fails** (red) in CI whenever it is selected but the source-tree `build/trtmc` binary is not present.
It should **skip** — a missing build artifact is an environment prerequisite, not a product defect.

## Symptom / repro

Observed repeatedly (3/3 identical) on Premerge CI for a `shared_builder_module` change
([run 26924196563](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/26924196563),
which currently blocks PR #196):

```
def test_time_series_models_match_reference_end_to_end():
>       assert TRTMC_BIN.exists(), f"Expected built binary at {TRTMC_BIN}"
E       AssertionError: Expected built binary at .../build/trtmc
tests/engine_defs/torch_trt/test_time_series_e2e.py:112: AssertionError
```

## Why it only bites some PRs

The test runs only under **broad tier-level** test selection — impact analysis logs
`No coverage map available -- using tier-level selection`, which a change to a
`shared_builder_module` (e.g. `engine_builder.py`) triggers. Most PRs get **narrow**
selection and never run this test, so the bug is invisible — until a broad-selection PR
comes along, where it produces a **false red that blocks an unrelated change.**

## Root cause

1. **Hardcoded, non-overridable binary path.** The test pins
   `TRTMC_BIN = REPO_ROOT / "build" / "trtmc"`. Every other binary-using test resolves the
   path *overridably*:
   - the canonical e2e `trtmc_binary` fixture — `--trtmc-binary`, default `build/trtmc`,
     and **`pytest.skip` if missing** (`tests/e2e/conftest.py`);
   - `test_family_elf._find_trtmc_binary()` honors the **`TRTMC_BINARY`** env var;
   - CI runs e2e against the **wheel entrypoint on PATH** via
     `--trtmc-binary "$(command -v trtmc)"`. The binary *is* available in CI — this test
     simply can't be pointed at it.
2. **Hard `assert` instead of `pytest.skip`** for a missing prerequisite — inconsistent
   with the test's own `@requires_gpu` / `@requires_torchtrt` skip markers, and with its
   directory sibling `test_pixart_vs_hf.py`, which `pytest.skip`s when its binary is
   absent (that is why pixart never turns red).

The module docstring even states these tests are *"intentionally serial and tiny so they
can run in CI on a single GPU"* — CI execution was **intended**; the hardcoded path
defeats it.

## Proposed fix

Resolve the binary via `TRTMC_BINARY` (fallback to source-tree `build/trtmc`) and
`pytest.skip` when absent — matching the canonical fixture and the sibling test. To make
it **actually execute** in CI (rather than skip), set `TRTMC_BINARY=$(command -v trtmc)`
in the stage that runs `tests/engine_defs/torch_trt/`, exactly as the selective-e2e stage
already injects the installed entrypoint. That stage wiring is a CI-config choice for the
owner.

## Related

`tests/engine_defs/torch_trt/test_pixart_vs_hf.py` has the same brittle hardcode
(`"/workspace/tensorrt-model-connect/build/trtmc"`) but already skips, so it is not
breaking — recommend aligning it to `TRTMC_BINARY` in a follow-up.

samli-gpu:
## What

`tests/engine_defs/torch_trt/test_time_series_e2e.py` hard-asserted a source-tree
`build/trtmc` binary, so it **failed** (rather than skipped) anywhere that path is not
populated — including the CI wheel/conan flow, where `trtmc` is installed on `PATH`
instead. A missing build artifact is an environment gap, not a product defect, so it
should skip.

This change resolves the binary the way the rest of the suite already does:

- honor a **`TRTMC_BINARY`** override (as `test_family_elf._find_trtmc_binary` does),
  falling back to the source-tree `build/trtmc` for local dev;
- **`pytest.skip`** when the binary is not found — matching the canonical `trtmc_binary`
  e2e fixture and the directory sibling `test_pixart_vs_hf.py`, both of which already skip.

To make the test **actually execute** in CI (rather than skip), set
`TRTMC_BINARY=$(command -v trtmc)` in the stage that runs `tests/engine_defs/torch_trt/`,
exactly as the selective-e2e stage already injects the installed entrypoint — a CI-config
choice for the owner.

## Why

The test only runs under **broad tier-level** test selection (impact analysis with no
coverage map), which a `shared_builder_module` change triggers. There it produced a
**false red** that blocked an unrelated PR. Full root-cause analysis in #209.

## Validation

- `ruff check --config ruff.toml` on the changed file: passes.
- Binary absent → the test now **skips** cleanly instead of failing:
  ```
  TRTMC_BINARY=/nonexistent/trtmc pytest tests/engine_defs/torch_trt/test_time_series_e2e.py
  SKIPPED [1] ...: trtmc binary not found: /nonexistent/trtmc (set TRTMC_BINARY ...)
  1 skipped
  ```
  (`@requires_torchtrt` / `@requires_gpu` did not skip it — torch-tensorrt and a GPU were
  present — so this is the new binary guard firing, not an unrelated skip.)

No behavior change when a binary is present: the test runs exactly as before.

Fixes #209

### DIFF
--- a/tests/engine_defs/torch_trt/test_time_series_e2e.py
+++ b/tests/engine_defs/torch_trt/test_time_series_e2e.py
@@ -23,7 +23,37 @@ transformers = pytest.importorskip("transformers")
 
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
-TRTMC_BIN = REPO_ROOT / "build" / "trtmc"
+
+
+def _resolve_trtmc_binary() -> Path:
+    """Resolve the C++ trtmc binary, honoring the ``TRTMC_BINARY`` override.
+
+    Mirrors the resolution used by the canonical ``trtmc_binary`` e2e fixture
+    and ``test_family_elf``: an explicit ``TRTMC_BINARY`` wins (e.g. the wheel
+    entrypoint CI installs, ``command -v trtmc``), otherwise fall back to the
+    source-tree ``build/trtmc`` used in local dev.
+    """
+    env_path = os.environ.get("TRTMC_BINARY")
+    if env_path:
+        return Path(env_path)
+    return REPO_ROOT / "build" / "trtmc"
+
+
+TRTMC_BIN = _resolve_trtmc_binary()
+
+
+def _require_trtmc_binary() -> None:
+    """Skip (don't fail) when the trtmc binary prerequisite is unavailable.
+
+    A missing build artifact is an environment gap, not a product defect, so it
+    is gated like the ``@requires_gpu`` / ``@requires_torchtrt`` prerequisites
+    above and like the sibling ``test_pixart_vs_hf`` test, which also skips.
+    """
+    if not TRTMC_BIN.is_file():
+        pytest.skip(
+            f"trtmc binary not found: {TRTMC_BIN} "
+            "(set TRTMC_BINARY to the installed entrypoint, or build ./build/trtmc)"
+        )
 
 
 def _has_torchtrt() -> bool:
@@ -72,7 +102,7 @@ def _parse_solve_stdout(stdout: str) -> torch.Tensor:
 
 
 def _build_bundle(model_dir: Path, bundle_path: Path, *, max_cache_length: int) -> None:
-    assert TRTMC_BIN.exists(), f"Expected built trtmc binary at {TRTMC_BIN}"
+    _require_trtmc_binary()
     _run(
         [
             str(TRTMC_BIN),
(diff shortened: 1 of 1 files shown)


samli-gpu:
Superseded by https://github.com/NVIDIA/TensorRT-Model-Connect/pull/214 — recreated from a branch in NVIDIA/TensorRT-Model-Connect so premerge CI can run (the CI router only accepts same-repo branches, not forks). Same commit and diff.

PR #242, opened 2026-06-12
fix(tests): drop unsettable ITensor.dtype in graph-op test harness (TRT 11)

samli-gpu:
## What

`tests/builder/conftest.py::run_trt_graph` (the shared helper behind the graph-op
GPU tests) builds a **strongly-typed** TRT network and then does:
```python
network.mark_output(tensor)
tensor.dtype = trt.float32
```
In a strongly-typed network TRT **infers** all tensor types and `ITensor.dtype`
is not meant to be set — and the `ITensor.dtype` **setter was removed in TRT 11**.
On TRT-11 runners that line raises:
```
AttributeError: property of 'ITensor' object has no setter
tests/builder/conftest.py:163: AttributeError
```
which fails ~every graph-op test at once (one run: 122 failures across
`test_graph_ops.py`, `test_graph_ops_extended.py`, `test_graph_blocks.py`).

The set is **redundant** — these test graphs are fp32, so the outputs are already
fp32 — so this drops it.

## Why it surfaced now
The runner pool is migrating to **TRT 11** (`remove-torchtrt-trt11`, `trt11-*`).
The stage passes on TRT-10 runners (setter still present) and fails on TRT-11
runners — so it's intermittent today and a hard block as the migration completes.
Because the **Graph-op GPU** stage runs before **Selective E2E**, a failure here
skips Selective E2E entirely, blocking unrelated PRs from getting an e2e result.

## Validation
- `ruff check --config ruff.toml tests/builder/conftest.py`: passes.
- Ran the full graph-op suite locally (TRT 10.16) with and without this change:
  **identical results** — the only failures are 2 pre-existing `TestAddCausalPad1d`
  cases that fail the same way on the original `conftest.py` (a separate, local
  TRT-op issue, not related to output dtype). Removing the line changes nothing on
  TRT 10 and removes the TRT-11-incompatible setter call.
- Only occurrence of the pattern in the repo (`grep` for `.dtype = trt.`).

(Validated on TRT 10 locally; the TRT-11 behavior is the removed `ITensor.dtype`
setter — confirmed against the CI symptom and the strongly-typed-network design.)

### DIFF
--- a/tests/builder/conftest.py
+++ b/tests/builder/conftest.py
@@ -159,11 +159,10 @@ def run_trt_graph(build_fn, inputs: dict[str, np.ndarray]) -> dict[str, np.ndarr
 
     for name, tensor in trt_outputs.items():
         tensor.name = name
+        # Strongly-typed network: output dtypes are inferred from the graph and
+        # are not settable (the ITensor.dtype setter was removed in TRT 11).
+        # These test graphs are fp32, so no explicit output cast is needed.
         network.mark_output(tensor)
-        # In newer TRT (10.x), `ITensor.dtype` is read-only after the layer
-        # graph is built; output dtype is inferred from the producing layer
-        # (here fp32 because inputs + weights are fp32). The explicit set
-        # was a no-op on older TRT and broke on newer.
 
     plan = builder.build_serialized_network(network, config)
     if plan is None:


samli-gpu:
Closing as redundant — `main` already removed `tensor.dtype = trt.float32` from tests/builder/conftest.py (came in via #237, 2026-06-12), which is exactly this fix. This PR only reworded the accompanying comment. The graph-op GPU stage is already unblocked on main; PRs just need to be tested against current main (their merge ref inherits the fix).

PR #262, opened 2026-06-25
fix(runtime): size vision-language KV cache from the engine dtype

pratacha-nv:
The vision_language KvCache sized its key/value buffers from the --precision string (e.g. bf16 -> 2 bytes per element), but the qwen_vl decoder keeps its KV tensors in fp32 regardless of --precision. With a bf16 or fp16 base the engine's fp32 present_k/present_v writes then overflowed the undersized buffers and corrupted the heap: usually garbage output, intermittently a Myelin l2cm illegal-memory-access crash.

Read the cache element dtype from the engine via
TrtModule::tensor_dtype(cache_k_name) instead of inferring it from the precision string, so the buffers always match what the engine writes. This unblocks reduced-precision (bf16/fp16) vision-language decoding.


pratacha-nv:
Superseded: ToT already sizes the VL KV cache from the engine dtype upstream. The rebased VL-precision work (#430) declares the KV inputs/outputs at the work dtype so the runtime buffer matches decode precision; no separate runtime change needed. Closing.

PR #263, opened 2026-06-25
fix(quant): int8 SmoothQuant and fp8-on-fp16-base accumulation

pratacha-nv:
## What
Two quant-format correctness fixes:
- **int8 SmoothQuant** — correct per-channel weight scales and apply SmoothQuant smoothing (migrate the per-input-channel activation range into the weights via `pre_quant_scale`) so the graph matches the ModelOpt-calibrated (smoothed-weight) scales. Prior int8 was per-tensor-only and mismatched the calibrated scales.
- **fp8 accumulation** — accumulate the fp8 matmul in fp32 on an fp16 base for numerical stability.

## Validation (rebased on ToT · 805 MMMU-Pro · calib-512 real manifest)
int8 **37.9%** (ref master 42.0%), garbage=0. The residual vs master is the shared upstream vision-encoder delta (see stack note), not this change — int8 tracks its expected drop from bf16.

Part of the VL quant/precision stack: **#263 → #264 → #430 → #431 → #390**.

### DIFF
--- a/python/tensorrt_model_connect/fp8_calibrate.py
+++ b/python/tensorrt_model_connect/fp8_calibrate.py
@@ -23,6 +23,8 @@ import re
 import sys
 from typing import Callable, TYPE_CHECKING
 
+import numpy as np
+
 if TYPE_CHECKING:
     import torch.nn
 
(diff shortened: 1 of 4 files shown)


pratacha-nv:
Content merged into main via #390 (rebase-and-merge, 2026-07-31) — commits 6c38908d/65f4a4b1/1f846329/2e43b90a/7daa02ad/50ec3b6d. Closing this stale branch (rebase rewrote SHAs so it didn't auto-close).

PR #281, opened 2026-07-01
[codex] test(ci): red-team model isolation

yifeif-nv:
## Background

This is a disposable red-team PR for the model-family source-projection boundary. It intentionally makes the GPT-2 builder depend on a BERT-owned resource while avoiding a direct sibling Python import.

The full checkout still contains BERT's `MODEL.toml`, so ordinary builder validation remains green. The strict GPT-2 source projection removes BERT, and CI is expected to fail when the GPT-2 plugin loads.

## Exit Criteria

- Full-checkout architecture and GPT-2 builder tests pass.
- Selective CI attributes the change to the GPT-2 family.
- The model-owned isolation rerun stages only GPT-2 and fails on the missing BERT resource.
- The PR remains draft and is not merged.

## Implementation

- Add an import-time GPT-2 validation that reads BERT's family metadata.
- Construct the sibling name dynamically so this exercises source projection rather than the existing direct-import AST guard.

## Validation

- `pytest tests/tools/test_model_plugin_encapsulation_static.py -q`: 155 passed.
- `pytest tests/tools/test_model_plugin_isolation.py -q`: 42 passed.
- Full checkout GPT-2 TP dispatch test: 1 passed.
- Projected GPT-2 TP dispatch test: expected failure with `FileNotFoundError` for `families/bert/MODEL.toml`.
- Impact analysis selects `distilgpt2` and `gpt2-125m` for L0 E2E.

## Notes

Expected status: red. This PR is evidence for the CI guard and must be closed after the expected isolation failure is captured.

### DIFF
--- a/.github/scripts/run-trtmc-ci.sh
+++ b/.github/scripts/run-trtmc-ci.sh
@@ -896,7 +896,13 @@ run_isolated_e2e_group_logged() {
     fi
   else
     echo "FAIL isolated group=$group_id gpu=$gpu_id rc=$group_rc elapsed=${elapsed}s" >&2
-    tail -n 120 "$audit_dir/console.log" >&2 || true
+    if ! awk '
+      /^=== Isolation result verification ===$/ { emit = 1 }
+      emit { print }
+      END { if (!emit) exit 1 }
+    ' "$audit_dir/console.log" >&2; then
+      tail -n 80 "$audit_dir/console.log" >&2 || true
+    fi
   fi
   return "$group_rc"
 }
--- a/.github/scripts/run-trtmc-ci.sh
+++ b/.github/scripts/run-trtmc-ci.sh
@@ -828,7 +828,7 @@ PY
       PYTHONPATH="$source_dir/python:$source_dir" \
       LD_LIBRARY_PATH="$isolated_library_path" \
       "${HF_PYTHON:-/opt/venv/bin/python}" -m pytest \
-        "${group_test_files[@]}" -v \
+        "${group_test_files[@]}" -v --tb=short \
         --rootdir "$source_dir" \
         -c "$source_dir/pyproject.toml" \
         --engine-dir "$engine_dir" \
@@ -848,6 +848,7 @@ PY
   popd >/dev/null
 
   local verification_rc=0
+  echo "=== Isolation result verification ==="
   if python3 "$source_dir/tools/model_plugin_isolation.py" verify-results \
       --repo-root "$source_dir" \
       --models-file "$models_file" \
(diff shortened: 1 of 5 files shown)


yifeif-nv:
## Red-team CI result

Expected failure confirmed in [run 28530552300](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/28530552300).

- Every required step through Graph-op GPU tests passed.
- Full-checkout selective E2E passed both `distilgpt2` and `gpt2-125m`.
- Strict isolation staged 739 files for only the GPT-2 builder family, E2E family, and runtime plugin.
- The isolated CMake build completed and linked exactly `libtrtmc_model_gpt2.so`.
- Both isolated bundle builds then failed with `FileNotFoundError` for the excluded `python/tensorrt_model_connect/families/bert/MODEL.toml`.
- Result verification rejected both cases as `build_fail`, causing `Selective E2E tests` and the workflow to fail.

This confirms the source-projection CI catches a filesystem/resource dependency on another model family even when the existing sibling-import AST guard does not. Keep this PR draft; do not merge.

yifeif-nv:
## Isolation failure output improved

Verified on [run 28533108982](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/28533108982) at commit `a5b01122`.

- All required gates through Graph-op GPU tests passed.
- Full-checkout E2E again passed `distilgpt2` and `gpt2-125m`.
- Strict isolation failed as expected.
- The failed-group output now uses a short pytest traceback and then prints only the structured verification section.
- Each model is labeled `build_fail` and shows the actual missing `families/bert/MODEL.toml` path under `Root cause`.
- The previous repeated CMake output and full runner implementation traceback are no longer emitted by the failed-group wrapper.
- Failures before result verification still use a bounded 80-line raw-log fallback.

The PR remains intentionally red and draft.

yifeif-nv:
## Newcomer-facing isolation explanation

Commit `b2265f09` now classifies missing resources against the model ownership roots before printing the low-level command error.

For this red-team case the failure begins with:

```text
MODEL ISOLATION VIOLATION DETECTED
  CI rebuilds and runs each model family from a filtered source tree that contains only that family's owned files plus shared/plugin infrastructure.
  Selected model family (builder files): 'gpt2'
  Resource-owning model family (builder files): 'bert'
  Cross-family resource: python/tensorrt_model_connect/families/bert/MODEL.toml
  Why CI failed: 'gpt2' used a resource owned by 'bert'. Strict isolation intentionally removes resources owned by other model families, so cross-family dependencies fail.
  Required boundary: keep the dependency in the selected model family's owned files or move it to approved shared/plugin infrastructure.
```

The structured classification is also persisted in `verification.json` as `isolation_violations`.

Local replay against the real failed CI artifacts produced this exact explanation. Final GitHub run [28534073571](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/28534073571) is queued with `steps: []` behind the broad selective E2E for PR #238; it will run automatically when the single self-hosted runner is released.

PR #533, opened 2026-07-23
ci(nightly): run all validation stages

yifeif-nv:
## Background

Scheduled Nightly run [29913059891](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/29913059891) failed in `Coverage and wheel package`. Although both Nightly matrices already used `fail-fast: false`, job-level success predicates skipped all 77 model proofs and the diffusion assessment. The final report and issue tracker ran, but the issue could only report the coverage failure and the downstream skips.

## Exit Criteria

- An upstream Legal, inventory-validation, source-quality, unit, cache, or package failure does not suppress independent Nightly validation.
- Every model matrix sibling runs after ordinary failures, and diffusion assessment is attempted after model failures.
- The combined report, final gate, and official scheduled failure-issue reconciliation still run and report the complete available evidence.
- Failed validation remains red and cannot publish Nightly wheels.

## Implementation

- Replace upstream-success predicates with `!cancelled()` so ordinary failures do not cascade while explicit cancellation still stops expensive work.
- Keep every existing dependency edge so cache/unit/package work finishes and releases runner capacity before the full model wave.
- Reuse the legal workflow's immutable SHA when available and fall back to the immutable trigger SHA after a legal failure. Promotion still requires the legal-certified SHA and a successful final gate.
- Let inventory selection continue after manifest validation fails so a valid matrix can still drive diagnostics. A checkout or selection failure that cannot produce a matrix remains the necessary hard-data exception.
- Add workflow contracts that forbid Nightly fail-fast, enforce the SHA fallback boundary, preserve success-only release, and verify the scheduled issue path fails loudly if its token is unavailable.

## Validation

- `python3 -m pytest tests/tools/test_github_actions_ci.py tests/tools/test_nightly_issue_tracker.py -q`: 114 passed.
- `ruff check tests/tools/test_github_actions_ci.py tests/tools/test_nightly_issue_tracker.py`: passed.
- `git diff --check`: passed.
- YAML safe-load check: passed and discovered all 12 Nightly jobs.
- Broader `tests/tools` attempt: 1,678 passed; 18 unrelated `test_cpu_profile.py` tests require the CUDA Python module missing from this host environment.

## Notes For Future Readers

- Review `.github/workflows/nightly.yml` first, then the contract updates in `tests/tools/test_github_actions_ci.py`.
- Publication is intentionally not exhaustive: `release` remains gated on complete Nightly success.
- Dynamic model dispatch cannot run if inventory checkout/selection produces no matrix. Explicit cancellation, GitHub platform outages, and issue API outages also remain external boundaries.
- Manual `workflow_dispatch` runs never mutate the production Nightly failure issue.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -44,7 +44,7 @@ jobs:
   inventory:
     name: 2 / All-model inventory
     needs: legal
-    if: ${{ needs.legal.result == 'success' }}
+    if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:
@@ -59,22 +59,24 @@ jobs:
       mode: ${{ steps.inventory.outputs.mode }}
 
     steps:
-      - name: Check out certified nightly snapshot
+      - name: Check out exact nightly snapshot
+        id: checkout
         uses: actions/checkout@v4
         with:
-          ref: ${{ needs.legal.outputs.tested_sha }}
+          ref: ${{ needs.legal.outputs.tested_sha || github.sha }}
           fetch-depth: 0
           persist-credentials: false
 
       - name: Validate ownership manifests
         env:
-          TESTED_SHA: ${{ needs.legal.outputs.tested_sha }}
+          TESTED_SHA: ${{ needs.legal.outputs.tested_sha || github.sha }}
         run: python3 tools/model_ci.py validate --revision "$TESTED_SHA"
 
       - name: Select every ownership model
         id: inventory
+        if: ${{ !cancelled() && steps.checkout.outcome == 'success' }}
         env:
-          TESTED_SHA: ${{ needs.legal.outputs.tested_sha }}
+          TESTED_SHA: ${{ needs.legal.outputs.tested_sha || github.sha }}
         run: |
           set -euo pipefail
           python3 tools/model_ci.py all \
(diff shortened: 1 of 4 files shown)


yifeif-nv:
Live validation evidence

Exact-head Pre-Merge

- PR head 7d12d3e553f3a078f94e993f89d0ecb14b7c449a passed run 29983074627: 11 successful jobs, one expected no-model skip, zero failures/cancellations.
- All five fallback model proofs, the combined report, and the required gate passed.

Negative Nightly canary

- Run 29985180288 used test-only failures for Legal, Package, and bark on top of the PR workflow.
- Inventory, source quality, unit, both per-node cache warm jobs, and all 77 model jobs still ran.
- The matrix completed 73 success / 4 failure / 0 cancelled / 0 skipped. All 28 generic runners were used: compute01 16/16 and compute02 12/12.
- The three non-injected failures were preserved as independent diagnostics: qwen3_omni transient TensorRT builder failure, magpie_tts threshold-sensitive reference ASR result, and patchtst GPU cleanup confirmation timeout. None cancelled a sibling.
- Diffusion semantic assessment ran after the failed model matrix and succeeded.
- Combined report ran, uploaded trtmc-nightly-html-report-29985180288-1, discovered all 77 model artifacts, and failed closed with the complete available evidence.
- The final Nightly gate ran and failed. Publication skipped as required.
- The failure-issue job skipped only because this was a manual non-main canary; official scheduled main runs retain the always-run issue reconciliation path.

This proves ordinary upstream and matrix failures no longer cascade into downstream skips while cancellation, hard missing matrix data, success-only publication, and production-only issue mutation remain explicit boundaries.

PR #628, opened 2026-07-27
feat(llm): add native fixed-capacity KV cache

yifeif-nv:
## Background

The native Qwen and Llama decoder paths treated KV-cache capacity as a user-selected build-time profile. That makes the build artifact unnecessarily deployment-specific and prevents the model-only flow we want: select a model at build time, then let its native runtime own the cache allocation.

This prototype qualifies two concrete models:

- `Qwen/Qwen3-0.6B`
- `nvidia/Llama-3.1-Minitron-4B-Depth-Base`

## Exit Criteria

- A model-only build selects the qualified model's configured sequence capacity without a public cache-length flag.
- Split prefill and decode engines share one runtime-owned full-capacity KV state.
- Attention work follows the active prefix rather than the physical capacity.
- The native path uses TensorRT APIs only; no custom attention plugin or kernel is added.
- Edge-LLM adapter behavior remains unchanged for explicit legacy deployment options.

## Implementation

- Build static full-capacity cache tensors and update them with TensorRT `IKVCacheUpdateLayer`.
- Run fused causal attention through TensorRT `IAttentionV2`, using `key_value_lengths` to bound the active prefix.
- Alias every `present_k/v` output to its corresponding `cache_k/v` input.
- Allocate one complete KV state when the pipeline loads and bind the same pointers to both split engines.
- Reject `kv_cache_size_bytes` for these native bundles instead of silently allocating a partial state.
- Keep the default-route policy model-owned and limited to the two qualified model configurations.
- Extend the shared TensorRT module backend with explicit alias ownership and external-binding propagation.

## Validation

- `PYTHONPATH=python CUDA_VISIBLE_DEVICES=1 /opt/venv/bin/python -m pytest -q tests/builder/test_cli.py tests/builder/test_optimized_runtime_public_routing.py tests/e2e/models/qwen/test_qwen_family_plugin_weights.py tests/e2e/models/qwen/test_qwen_builder_engine.py tests/e2e/models/llama/test_llama_family_plugin_weights.py tests/e2e/models/llama/test_llama_standard_decoder.py`: **104 passed**
- `ctest --test-dir build-native-kv --output-on-failure -R "test_trt_module|test_qwen_native_kv_cache|test_llama_native_kv_cache|test_qwen_tensor_names|test_llama_plugin_helpers|test_llama_pipeline"`: **6 passed**
- Qwen3-0.6B real bundle: split prefill/decode enqueue passed with 40,960 physical rows and one 4.38 GiB KV state.
- Llama-3.1-Minitron-4B real bundle: split prefill/decode enqueue passed with 131,072 physical rows and one 8.00 GiB KV state; 12 generated tokens matched the HF BF16 reference.
- Qwen full-capacity and legacy 256-row short-decode measurements were both approximately 1.55 ms/token on the qualified GB300 environment.
- A native Qwen run with `--kv-cache-size 1G` failed with the expected fixed-capacity diagnostic.
- `git diff --check`: passed.

## Notes For Future Readers

- Suggested review order: model graph construction, shared TensorRT alias handling, model runtime allocation/binding, then default routing and tests.
- This proves full static shapes, actual full-buffer allocation, shared split-engine execution, and short-sequence model runs. It does not yet include a near-40K or near-131K end-to-end soak.
- Qwen's 40,960 value is the checkpoint's total position/KV capacity; it should not be described as a 40K native input window.
- The split bundles contain two engine plans, so plan storage is larger even though runtime KV state is allocated only once.


yifeif-nv:
Superseded by #673, which now contains the consolidated Qwen and Llama full-context native KV implementation on current main with a KISS-reduced two-commit history. Closing this review surface so all review and CI continues in #673.

PR #834, opened 2026-08-06
test(minimax-h3): enforce coverage contracts

yifeif-nv:
## Context

The historical QA dashboard run `trtmc-validate-gb300-2-20260726-c8291be0-all105` was produced from Source revision `c8291be0390705ca1cdcccdff83e251e7d7e1767` and contains 105 rows. It predates MiniMax-H3, so this PR does **not** claim that MiniMax was one of that run's failing models. MiniMax-H3 was added to `main` later by #815; the missing contracts below then became a shared current-`main` blocker for the rebased model-accuracy PRs.

MiniMax-H3 onboarding was incomplete in four repository-wide contracts:

- the validation catalog did not contain `minimax-h3-768p`;
- release-performance accounting neither ran nor explicitly excluded it;
- a package-registry assertion inside the family source violated family isolation; and
- the strict family inventory still expected 78 families after the repository had 79.

The model-owned premerge lane selected `tests/builder/`, while the canonical coverage, performance, isolation, and inventory assertions lived under `tests/tools/`. That topology let a model land without those contracts being exercised by its normal premerge projection.

The first validation registration also exposed an execution cycle: a clean-engine reference tried to obtain Source provenance from a bundle before that bundle existed. Exact-head model proof then exposed a separate resource-contract gap: `exclusive_gpu` acquired cooperative lock slots but, with no free-memory requirement, could start on a device without enough capacity to deserialize the first TensorRT plan.

## What changed

- Added a one-sample, model-owned validation profile for the existing pinned `minimax-h3-768p` contract: checkpoint `48d93ede732756e404a3b1b2f3b3a9b5a22f6cfc`, Diffusers reference `abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc`, the manifest prompt, seed 0, 1344x768, 124 frames, 50 steps, 24 fps, and BF16 candidate/reference execution.
- Resolved `reference_source_revision: current` to an exact lowercase Source SHA before reference-cache preparation. The same SHA is part of the reference-cache identity, is passed into a clean bundle build, and is verified on both bundle reuse and after a rebuild. Missing or mismatched provenance fails closed.
- Added an explicit MiniMax release-performance exclusion until the pinned Diffusers reference is integrated into the release runner. This records the unsupported path; it does not relax a timing or accuracy threshold.
- Moved the package-registry assertion out of the family source and kept the family-isolation allowlist unchanged.
- Added builder-owned wrappers for validation coverage, performance coverage/exclusion, family source isolation, and the strict 79-family inventory.
- Classified a declarative `tests/performance/release.yaml`-only change as source-contract-only. Changes to selector/runtime/performance-runtime code retain the fixed fallback set.
- Declared `exclusive_gpu` plus `184320 MiB` of required free capacity for MiniMax. The value is based on the successful identical HF reference's measured `144177.109 MiB` peak plus about 39.2 GiB of headroom. Admission also rejects active compute processes. The manifest value and selector propagation are both locked by tests.

No visual comparison threshold was added, removed, or weakened.

## Reproduction and investigation

- The initial exact-head source lane caught the four current-`main` onboarding omissions once the builder wrappers were present, including the stale 78-versus-79 family inventory.
- A clean-engine run showed that reading provenance from a not-yet-built bundle made the validation row non-runnable. Resolving the exact Source SHA before cache/build preparation removes that cycle without bypassing provenance checks.
- On head `ce933d8857d2e22ca9160bacacd0532af9e9f23c`, one model-proof attempt failed during TensorRT build; the next built a 121 GB bundle and then failed while deserializing the first native plan because the manifest had `min_free_gpu_memory_mib=0`. Native inference, HF reference, and comparison correctly did not run after that failure.
- The stale `82d5b025d2ff39bb1b1135c3c82de38184baa33d` diagnostic run returned `rc=1` from the TensorRT text-encoder builder without an OOM, SIGKILL, disk-pressure, or deeper TensorRT error in its artifact. It was superseded by the current head and is not used as merge evidence.
- The current exact head passed from a clean scratch build through native execution, pinned HF reference, and all-frame comparison. That result, not a stale or partially completed run, is the readiness evidence below.

## Exact-head validation

Source head: `529e2e0210f29b0eaa0867d110c5051da834f780`

- [Private premerge run 31089031941](https://github.com/NVIDIA-dev/TensorRT-Model-Connect-CI/actions/runs/31089031941): **PASS**
  - source-only Python: `3036 passed, 1 skipped`;
  - projected Python: `17 passed, 128 deselected`;
  - source-only C++: `38/38 passed`;
  - MiniMax family Python: `88 passed`;
  - MiniMax E2E: `16 passed in 1505.22s`;
  - DeepSeek-V2, LTX-Video, Whisper, PatchTSMixer, and SegFormer fixed fallbacks: all passed.
- [MiniMax artifact 8963534728](https://github.com/NVIDIA-dev/TensorRT-Model-Connect-CI/actions/runs/31089031941/artifacts/8963534728): exact-head receipt verified
  - exclusive GPU 0, all four lease slots, no network access;
  - capacity admission: `184320 MiB` required, `280607 MiB` effective free observed;
  - clean bundle build: one invocation, one attempt, `rc=0`, 843.27s, 121,486,206,616 bytes;
  - native and HF receipts both bind Source `529e2e...` and checkpoint `48d93e...`;
  - both produced exactly 124 finite RGB frames at 1344x768;
  - minimum/mean low-frequency correlation: `0.907871 / 0.939423` against `>= 0.8 / >= 0.9`;
  - brightness-profile correlation: `0.999868` against `>= 0.95`;
  - maximum frame-brightness absolute error: `0.013277` against `<= 0.08`;
  - temporal-activity correlation: `0.999447` against `>= 0.9`;
  - maximum temporal-activity absolute error: `0.007833` against `<= 0.05`;
  - frame-contrast ratios remained inside `[0.7, 1.4]`;
  - PSNR `19.4489 dB` and MAE `0.04554436` remain diagnostic, as defined by the unchanged model contract.
- [CodeQL run 31088989028](https://github.com/NVIDIA/TensorRT-Model-Connect/actions/runs/31088989028): JavaScript/TypeScript and Python both passed.
- `trtmc/premerge/required`: exact-head success, published from the private verdict above.
- Ruff and `git diff --check`: passed.

The current PR changes shared selection contracts, so its one-time exact-head run intentionally selected one direct MiniMax proof plus five fixed fallback models. The full private run completed in about 34 minutes; it did not fan out to all models.

Steady-state cost is bounded:

- the added builder wrappers are source-only checks;
- a future `release.yaml`-only change runs unit/all with zero GPU model proofs;
- a MiniMax-owned change selects one direct MiniMax proof; and
- shared selector/runtime or performance-runtime changes retain the existing fixed five-model fallback rather than an all-model fanout.

The GPU capacity check adds no model execution. It waits for a suitable exclusive device before the expensive build, avoiding a late native-start OOM after spending roughly 14-20 minutes building the bundle.

## Follow-up boundary

Remove the release-performance exclusion only after the pinned Diffusers checkout is wired into the release runner and validated on target hardware. Do not replace exact-source binding with a cache bypass, move the registry assertion back into the family package, or weaken the model-owned visual thresholds.

### DIFF
--- a/benchmarks/performance/release.yaml
+++ b/benchmarks/performance/release.yaml
@@ -17,6 +17,12 @@ defaults:
     input_preparation_included: true
     asset_loading_included: false
 
+excluded_profiles:
+  - model: minimax-h3-768p
+    reason: >-
+      The pinned Diffusers reference for MiniMax-H3 has not yet been integrated
+      into the release performance runner.
+
 entries:
   - id: albert.encode
     family: albert
--- a/python/tensorrt_model_connect/families/minimax_h3/tests/test_family.py
+++ b/python/tensorrt_model_connect/families/minimax_h3/tests/test_family.py
@@ -10,7 +10,6 @@ import tomllib
 
 import pytest
 
-from tensorrt_model_connect.families import find_diffusion_plugin, load_plugin_by_id
 from tensorrt_model_connect.families.minimax_h3.config import (
     DEFAULT_WORKSPACE_LIMIT_BYTES,
     MiniMaxH3Config,
@@ -65,14 +64,6 @@ def test_manifest_discovers_both_public_pipeline_names() -> None:
         "vae/**",
         "audio_vae/**",
     }
-
-
-def test_family_registry_loads_native_plugin_for_public_pipelines() -> None:
-    plugin = load_plugin_by_id("minimax_h3")
-    assert plugin is not None
-    assert plugin.name == "minimax_h3"
-    for pipeline_class in ("MiniMaxH3ModularPipeline", "MiniMaxH3Pipeline"):
-        assert find_diffusion_plugin(pipeline_class) is plugin
 
 
 def test_plugin_aliases_are_exact() -> None:
--- a/tests/builder/test_repository_contracts.py
+++ b/tests/builder/test_repository_contracts.py
@@ -6,6 +6,9 @@
 from tests.tools.test_family_source_isolation import (
     test_family_imports_resolve_without_sibling_or_unapproved_shared_modules as _check_family_isolation,
 )
+from tests.tools.test_family_specialization import (
+    test_repository_registers_all_current_families as _check_family_inventory,
+)
 from tests.tools.test_perf_matrix import (
     test_release_suite_covers_every_non_l0_ready_model_profile as _check_release_coverage,
 )
@@ -24,3 +27,7 @@ def test_ready_models_have_release_performance_coverage_or_exclusion() -> None:
 
 def test_family_sources_remain_isolated() -> None:
     _check_family_isolation()
+
+
+def test_repository_family_inventory_is_current() -> None:
+    _check_family_inventory()
--- a/tests/builder/test_repository_contracts.py
+++ b/tests/builder/test_repository_contracts.py
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Run repository-wide ready-model contracts in every builder lane."""
+
+from tests.tools.test_family_source_isolation import (
+    test_family_imports_resolve_without_sibling_or_unapproved_shared_modules as _check_family_isolation,
+)
+from tests.tools.test_perf_matrix import (
+    test_release_suite_covers_every_non_l0_ready_model_profile as _check_release_coverage,
+)
+from tests.tools.test_trtmc_validate import (
+    test_model_workload_catalog_covers_every_ready_model as _check_validation_coverage,
+)
+
+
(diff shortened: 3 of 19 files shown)


chaofengw-nv:
One additional CI correctness gap surfaced while reviewing this work:

The required premerge result is currently tied to the PR head tested with the main revision captured when the run started. If main advances before the PR is merged, GitHub may still accept that older result even though the final combination of latest main plus this PR was never tested. The PR itself does not need to change for that combination to become different.

Could you evaluate whether strict required status checks or a merge queue are feasible here? A simple target flow would be: create a temporary merge commit from the latest main plus the PR, run the required checks against that exact merge SHA, report the required status on that SHA, and discard or cancel the result if main advances. Starting with a merge-queue batch size of one would keep the behavior and GPU cost easier to control.

PR #856, opened 2026-08-10
fix(python): complete model execution profiles

chaofengw-nv:
## Background

Accuracy and performance qualification creates model-owned Python profiles instead of relying on ambient packages from the runner image. Clean Thor deployments exposed missing or incompatible dependencies in several build and reference profiles.

## Exit Criteria

- Affected profiles declare and verify their complete dependency closure.
- Nested profiles retain inherited site-packages without leaking unrelated host packages.
- Architecture-specific dependencies use exact, reviewable platform markers.
- OpenCLIP, DeepSeek MoE, and timm references execute from self-contained profiles.
- Bundle construction resolves the selected build profile instead of relying on ambient Python.
- This PR owns every source input that changes the shared CI profile-image fingerprint.

## Implementation

- Complete the family-owned environments and verification contracts introduced by this PR.
- Pin the OpenCLIP text dependency needed by clean reference profiles.
- Align the DeepSeek-V2 MoE package set and verify the selected experts implementation.
- Give timm its own reference profile instead of inheriting incompatible global packages.
- Complete Chronos forecasting's exact scikit-learn dependency closure and verify it during profile creation.
- Honor numeric `TORCH_CUDA_ARCH_LIST` targets even when CUDA extension sdists hard-code every toolkit architecture, avoiding redundant source builds and memory pressure.
- Include bootstrap locks in the image fingerprint and prebuild the selected Nemotron speech reference profile for prebuilt-only proof jobs.
- Resolve each model's build interpreter from its execution profile and prepend that interpreter to bundle-build subprocesses.

## Validation

- Current head `59158d07`: `PYTHONPATH=python python3 -m pytest -q tests/tools/test_e2e_python_profiles.py tests/tools/test_validation_engine.py`: 295 passed.
- Current head `59158d07`: Ruff over `tools/validation/engine.py` and `tests/tools/test_validation_engine.py`: passed.
- Current #856 and #870 heads resolve to the same CI image fingerprint, `1ee43aacbdc1c56a57200fab2e70ea3f2578b18c53df212427fed7a0b6aba82b`; #870 changes none of this head's fingerprint inputs.
- Earlier profile/timm selection at head `7608794d`: 132 tests passed.
- Full integration with #870 and the Thor follow-ups: 653 selected tests passed.
- Full image-profile materialization exposed Chronos' missing `scikit-learn`; its repaired profile materialized cleanly and the updated Chronos/profile contract suite passed (29 tests).
[...]

PR #870, opened 2026-08-13
fix(validation): harden L4T model checks

chaofengw-nv:
## Background

The platform-aware model-check entry point exposed L4T qualification gaps in resource preparation, runtime-library discovery, cache cleanup, device propagation, and platform admission. The L4T profile also over-excluded models by mixing successful runs and accuracy disagreements with the final not-compared set. Bare-metal L4T additionally keeps `nvcc` under the CUDA root without placing it on `PATH`, which caused CMake to silently omit model CUDA sources. A separately selected TensorRT runtime could also inherit stale Python bindings from the base environment and stamp bundles with the wrong ABI.

## Exit Criteria

- Accuracy and Perf can run serially from the unified entry point on a bare-metal TensorRT runtime.
- Generated engines and model caches follow the configured lifecycle policy while reports and diagnostic logs remain available.
- Passing models and accuracy disagreements remain eligible for L4T execution.
- The L4T model-wide exclusion list contains only the 11 agreed not-compared models.
- L4T state remains on `/dev/nvme0n1p1`.
- Standard CUDA-root installations compile model-owned CUDA kernels even when `nvcc` is absent from `PATH`.
- On-demand Python profiles inherit the TensorRT Python runtime selected by the execution environment.

## Implementation

- Prepare model resources on demand and preserve the selected runtime environment across child runners.
- Propagate runtime libraries, reference devices, and model-specific compatibility settings to worker processes.
- Keep model-wide platform exclusion separate from Accuracy workload selection.
- Restrict `l4t-thor` exclusions to the 11 consolidated not-compared models and cover the exact list with a unit test.
- Delete generated engines and per-model/per-entry HF caches according to platform retention policy without deleting diagnostic logs.
- Restore the Jetson L4T storage-device requirement to `/dev/nvme0n1p1` instead of applying the Thor X device layout.
- Preserve explicit `CUDACXX` selection and otherwise discover `nvcc` from configured or standard CUDA roots before enabling the CUDA language.
- Add `python_dirs` beside library and executable roots so a platform environment can prepend matching TensorRT Python bindings to every child and on-demand profile.

## Validation

- Current head `89356135`: `python3 -m pytest -q tests/tools/test_model_checks.py`: 55 passed.
- Current head `89356135`: `python3 -m pytest -q tests/tools/test_model_plugin_encapsulation_static.py`: 158 passed.
- Current head `89356135`: `ruff check tools/model_checks.py tests/tools/test_model_checks.py tests/tools/test_model_plugin_encapsulation_static.py`: passed.
- Current head `89356135`: `git diff --check`: passed.
- L4T target proof: a clean configure with a `PATH` that excludes `nvcc` selected `/usr/local/cuda/bin/nvcc`; all model plugins built 103 CUDA objects and `ldd -r` reported no unresolved model symbols.
- L4T target proof: configured Python and library roots made both the base runner and an existing model profile import TensorRT `11.2.1.2` instead of the installed `11.0.2.2` package.
- Parent head `83297d73`: focused model-check, validation, Perf, reference, Lance, Eagle VLM, and Wan tests: 642 passed.
- Current-head premerge is pending; no prior result is claimed for this head.

## Notes For Future Readers

- Review `tools/model_checks.py`, `tests/model_checks/platforms/l4t-thor.yaml`, `tests/model_checks/environments/l4t-thor.yaml`, and the CUDA discovery block at the start of `CMakeLists.txt` first.
- Python profile registries, locks, image inputs, and CI prebuild policy remain isolated in #856.
- Merge #872, then #856, before this PR. Target-platform qualification uses the combined changes, while this branch contains only #870-owned source changes.
- Accuracy disagreement is a retained test result, not a reason to exclude a model from L4T.


chaofengw-nv:
Lance SDPA follow-up is now included in head `2e0ee994`.

- Removes the tracked fake `flash_attn` distribution metadata.
- Selects SDPA through a run-owned VIT config overlay while leaving checkpoint files and shared Hugging Face caches unchanged.
- Keeps the Python compatibility module only for the direct Lance imports.
- Current-head selected regression suite: 678 passed. Ruff, diff validation, and `tools/legal_headers.py --check` passed.
- The equivalent pre-rebase patch revision completed Lance Accuracy at 5/5 with agreement 1.0 on both L4T Thor and GB300. This is target-hardware evidence for the patch, not an exact-current-head premerge claim.

Current-head CI is being requested separately.

PR #918, opened 2026-08-18
chore: withdraw unpublished change

yifeif-nv:
This pull request has been withdrawn and is no longer proposed for public review.


yifeif-nv:
Closing for an architectural reset. The current diff is too large and relies too heavily on vendored/custom CUDA. The replacement will start from current github/main, prefer TensorRT native operations throughout, retain only evidence-backed minimal custom plugins, and require exact-head accuracy and performance qualification before a new draft PR.

Issue #928, opened 2026-08-19
Feature: Add native LFM2 model family (LiquidAI hybrid conv+attention)


munnmajithia:
### Request type

New model or checkpoint support

### Problem and use case

LiquidAI's LFM2 family (350M/700M/1.2B/2.6B, `Lfm2ForCausalLM`, native in transformers >= 4.54, no `trust_remote_code`) is a widely used edge-focused text-generation line with a hybrid architecture: a majority of gated short-convolution layers with a small rolling conv cache (`conv_L_cache = 3`) plus a minority of GQA attention layers with QK-norm. TensorRT-Model-Connect has no `lfm2` family today, and no existing family resolves it. The hybrid families (`nemotron_h`, `qwen3_5`) own Mamba+attention semantics, not short-conv+attention. Users who want native TensorRT bundles for LFM2 checkpoints have no path.

### Proposed outcome

`trtmc build LiquidAI/LFM2-350M` produces a working bundle that passes continuation parity against the `hf_transformers` reference. Concretely, a three-sided model-owned unit per the Add-a-Model-Family guide:

- Python builder family `families/lfm2/` (checkpoint mapping for the conv/attention layer split, SwiGLU MLP, QK-norm, tied embeddings)
- Runtime DSO `src/runtime/models/lfm2/` with its own strategy key `lfm2_hybrid_conv_attention`, composing a family-local KV cache with a single-spec conv recurrent state (double-buffered `conv_state_{i}`/`present_conv_{i}` per the established CUDA-graph-safe pattern)
- E2E root `tests/e2e/models/lfm2/` with an `LFM2-350M` fp16 manifest as the core case, and the larger checkpoints (700M/1.2B/2.6B) as follow-up manifests

Out of scope for v1: LFM2-8B-A1B (MoE) and LFM2-VL, which have different architecture surfaces.

### Model and target details

`LiquidAI/LFM2-350M` (initial), task `text_generation_causal`, fp16, single consumer or datacenter GPU (validation planned on L4), runtime path: new model-owned `lfm2_hybrid_conv_attention` strategy.

### Alternatives considered

[...]

PR #942, opened 2026-08-20
fix(lfm2): match pinned pretokenizer whitespace splitting

munnmajithia:
## Background

The native LFM2 pretokenizer reimplements the pinned Split regex from the model's tokenizer.json. Its whitespace branch consumes newline-free whitespace runs whole, which drops the `\s+(?!\S)` alternative: a run followed by non-whitespace must yield its final whitespace unit to the following piece. Any prompt with two or more consecutive spaces before text therefore tokenizes differently than the pinned Hugging Face tokenizer, with indented code as the common case. Reported in #928. Scope is the native C++ tokenization path only, no crash, but token ids silently diverge from what the model was trained on.

| input | pinned HF tokenizer | before this PR |
|---|---|---|
| `a  b` | `a` &#124; ` ` &#124; ` b` | `a` &#124; `  ` &#124; `b` |
| `def f():\n    return  x` | `def` &#124; ` f` &#124; `():\n` &#124; `   ` &#124; ` return` &#124; ` ` &#124; ` x` | `def` &#124; ` f` &#124; `():\n` &#124; `    ` &#124; `return` &#124; `  ` &#124; `x` |

## Change

`whitespace_piece_end` yields the final whitespace unit of a newline-free run when the run is followed by non-whitespace. Single-unit runs and end-of-text runs are consumed whole, matching the plain `\s+` fallback. The newline branch (`\s*[\r\n]+`) is unchanged.

## Validation

- New `test_whitespace_run_piece_and_id_parity` in `test_lfm2_pretokenizer.cpp`: six fixtures (double space, space before digit, space plus tab, trailing run, indented code, markdown row) asserting both Split pieces and token ids. Ids are pinned from the LiquidAI/LFM2-350M tokenizer.json at manifest revision f37d3f5c (byte-identical to main), with piece-wise encoding verified equal to full-text encoding for every fixture. The five affected fixtures fail against the previous splitter (10 failed checks) and pass with the fix; the trailing-run fixture passes before and after, pinning the end-of-text behavior against over-correction.
- Corpus diff of `lfm2_split_pretokens` against Hugging Face tokenizers 0.23.1 running the pinned tokenizer.json: 424 cases covering targeted whitespace patterns, every existing multilingual fixture text, and 400 random whitespace-heavy fuzz strings. Before: 311/424 diverge from the reference pre-tokenization. After: 0/424.
- All existing fixtures pass unchanged.

Evidence environment: splitter and test compiled standalone with clang on a CPU-only macOS arm64 host, since the change touches no CUDA or TensorRT path. The registered `runtime_tests` entry is unchanged, so CI exercises the same test in the TensorRT/CUDA environment.

### DIFF
--- a/src/runtime/models/lfm2/pretokenizer.cpp
+++ b/src/runtime/models/lfm2/pretokenizer.cpp
@@ -229,15 +229,25 @@ std::size_t whitespace_piece_end(std::string_view text, std::size_t offset, cons
 
     std::size_t scan = offset;
     std::size_t last_newline_end = offset;
+    std::size_t last_unit_begin = offset;
     while (scan < text.size()) {
         const auto unit = decode_utf8_unit(text, scan);
         if (!is_whitespace(unit.codepoint))
             break;
+        last_unit_begin = scan;
         scan += unit.size;
         if (is_newline(unit.codepoint))
             last_newline_end = scan;
     }
-    return last_newline_end > offset ? last_newline_end : scan;
+    if (last_newline_end > offset)
+        return last_newline_end;
+    // \s+(?!\S): a newline-free run followed by non-whitespace yields its
+    // final whitespace unit to the next piece (it becomes that piece's
+    // prefix). A single-unit run has nothing left to yield and falls through
+    // to the plain \s+ alternative, consuming the unit whole.
+    if (scan < text.size() && last_unit_begin > offset)
+        return last_unit_begin;
+    return scan;
 }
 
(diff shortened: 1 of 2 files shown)


munnmajithia:
@yifeif-nv this is the fix from #928. CI trigger appreciated when you get a chance.

yifeif-nv:
> @yifeif-nv this is the fix from #928. CI trigger appreciated when you get a chance.

CI triggered

yifeif-nv (approved):

yifeif-nv:
Looks good! Merged

PR #981, opened 2026-08-21
fix(cli): unify kv-cache-size byte parser between CLI and benchmark

### LINKED ISSUE #971: [Refactor] Use one byte-size parser across CLI executables
## Parent

Inspired by #963: consolidate repeated parsing logic before its behavior drifts further.

## What to build

Provide one canonical byte-size parser for CLI-facing cache and memory options, and use it from both the main CLI and dataset benchmark. Define the accepted SI and IEC suffixes, rounding, overflow, whitespace, and error contract once.

The two current implementations already disagree: one treats KB/MB/GB as decimal units and KiB/MiB/GiB as binary units, while the other treats both spellings as binary units for the same `--kv-cache-size` option.

## Acceptance criteria

- [ ] The main CLI and dataset benchmark call the same byte-size parser.
- [ ] SI and IEC suffix behavior is documented and tested, including `1GB != 1GiB` if the existing main-CLI contract is retained.
- [ ] Case handling, fractional values, zero/negative values, trailing characters, rounding, and overflow have one explicit contract.
- [ ] The duplicate benchmark parser is removed.
- [ ] Cross-executable tests verify that identical option text produces identical bytes or identical validation errors.

## Blocked by

None - can start immediately.

lukiod:
Closes #971.

The dataset benchmark had its own `parse_size_bytes()`, which treated `KB`/`MB`/`GB` as 1024-based (binary) and threw `std::runtime_error` on error. The main CLI's `parse_byte_size()` treats `KB`/`MB`/`GB` as 1000-based (decimal, SI) and `KiB`/`MiB`/`GiB` as 1024-based (binary, IEC), returning `std::optional`. Same option (`--kv-cache-size`), same-looking input, different byte count depending on which executable parsed it.

## Fix

`trtmc_dataset_benchmark` now compiles and links `src/cli/args.cpp`, and calls the same `trtmc::cli::parse_byte_size()` the main CLI uses — matching the existing pattern `test_cli_args` already uses for the same source file. The benchmark's own `parse_size_bytes()` is now a thin adapter that preserves its existing throw-on-invalid contract at the one call site, rather than a second implementation of the parsing logic itself.

Kept the existing main-CLI contract (`1GB != 1GiB`) rather than changing it, per the issue's acceptance criteria.

## Tests

Added `test_parse_byte_size_si_vs_iec`, `test_parse_byte_size_case_and_fractional`, and `test_parse_byte_size_rejects_invalid` to `tests/cpp/test_cli_args.cpp` — there was no direct test coverage of `parse_byte_size()` before this, only indirect coverage via `test_bad_kv_cache_size_fails`. Covers SI vs IEC suffixes, case-insensitivity, fractional values, zero/negative rejection, trailing-garbage rejection, unsupported-suffix rejection, and overflow rejection.

Verified by compiling `tests/cpp/test_cli_args.cpp` + `src/cli/args.cpp` standalone (`g++ -std=c++17 -I src`) — all tests pass, including the new ones. Also verified the edited `examples/trtmc_dataset_benchmark.cpp` with a syntax-only compile (`-fsyntax-only`) against the project's real headers. I don't have a CUDA/TensorRT toolchain available to do a full linked build of `trtmc_dataset_benchmark` itself, so that specific executable's link step is unverified beyond the syntax check — flagging that directly rather than claiming a full build I didn't run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -691,7 +691,10 @@ set_target_properties(trtmc PROPERTIES
 )
 
 if(TRTMC_BUILD_BENCHMARKS)
-  add_executable(trtmc_dataset_benchmark examples/trtmc_dataset_benchmark.cpp)
+  add_executable(trtmc_dataset_benchmark
+    examples/trtmc_dataset_benchmark.cpp
+    src/cli/args.cpp
+  )
   target_link_libraries(trtmc_dataset_benchmark PRIVATE trtmc_core)
   target_include_directories(trtmc_dataset_benchmark PRIVATE ${PROJECT_SOURCE_DIR}/src)
   target_include_directories(trtmc_dataset_benchmark SYSTEM PRIVATE
--- a/examples/trtmc_dataset_benchmark.cpp
+++ b/examples/trtmc_dataset_benchmark.cpp
@@ -4,6 +4,7 @@
  */
 
 #include "cli/jsonl_io.h"
+#include "cli/args.h"
 #include "trtmc/config/cli_support.h"
 #include "trtmc/config/schema_registry.h"
 #include "trtmc/pipeline.h"
@@ -108,28 +109,10 @@ void usage() {
 }
 
 std::uint64_t parse_size_bytes(const std::string& text) {
-    if (text.empty())
-        throw std::runtime_error("Empty kv-cache-size");
-    std::size_t idx = 0;
-    const double value = std::stod(text, &idx);
-    std::string suffix = text.substr(idx);
-    for (char& ch : suffix)
-        ch = static_cast<char>(std::toupper(static_cast<unsigned char>(ch)));
-    double multiplier = 1.0;
-    if (suffix.empty() || suffix == "B") {
-        multiplier = 1.0;
-    } else if (suffix == "K" || suffix == "KB" || suffix == "KIB") {
-        multiplier = 1024.0;
-    } else if (suffix == "M" || suffix == "MB" || suffix == "MIB") {
-        multiplier = 1024.0 * 1024.0;
-    } else if (suffix == "G" || suffix == "GB" || suffix == "GIB") {
-        multiplier = 1024.0 * 1024.0 * 1024.0;
-    } else if (suffix == "T" || suffix == "TB" || suffix == "TIB") {
-        multiplier = 1024.0 * 1024.0 * 1024.0 * 1024.0;
-    } else {
-        throw std::runtime_error("Unsupported kv-cache-size suffix: " + suffix);
-    }
-    return static_cast<std::uint64_t>(value * multiplier);
+    auto parsed = trtmc::cli::parse_byte_size(text);
+    if (!parsed.has_value())
+        throw std::runtime_error("Invalid kv-cache-size: " + text);
+    return *parsed;
 }
 
 } // namespace
--- a/examples/trtmc_dataset_benchmark.cpp
+++ b/examples/trtmc_dataset_benchmark.cpp
@@ -3,8 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "cli/jsonl_io.h"
 #include "cli/args.h"
+#include "cli/jsonl_io.h"
 #include "trtmc/config/cli_support.h"
 #include "trtmc/config/schema_registry.h"
 #include "trtmc/pipeline.h"
--- a/tests/cpp/test_cli_args.cpp
+++ b/tests/cpp/test_cli_args.cpp
@@ -345,6 +345,40 @@ void test_bad_kv_cache_size_fails() {
     check(args.error_message.find("--kv-cache-size expects") == 0, "bad kv cache message");
 }
 
+void test_parse_byte_size_si_vs_iec() {
+    using trtmc::cli::parse_byte_size;
+    check(parse_byte_size("1GB") == 1000ull * 1000 * 1000, "1GB is decimal");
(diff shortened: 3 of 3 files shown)


chaofengw-nv:
Hi, thanks for the contribution, sharing the parser is the right approach. Could you please:
 1. Rebase onto the latest main so the improved contributor-facing CI can run.
 2. Add the cross-executable tests required by #971. The current tests only exercise parse_byte_size() directly, and invalid input still produces different errors in the two executables.
 3. Document and test the rounding and whitespace behavior, and reject non-finite inputs such as nan/inf.

It would also be helpful to include the complete validation commands and results. Thanks

lukiod:
Rebased onto latest main, and addressed all three points:

1. **Cross-executable divergence**: `trtmc_dataset_benchmark`'s own `parse_size_bytes()` wrapper is gone. Both it and the main CLI now call `parse_byte_size_or_throw()` (new, in `src/cli/args.h`/`.cpp`), and the error text itself is a single shared constant (`kInvalidByteSizeMessage`) used at all three call sites, so the message can't independently drift again. Also fixed a real crash: an invalid `--kv-cache-size` on the benchmark previously threw past an uncaught exception all the way out of `main()` (no try/catch anywhere in the parsing loop), which calls `std::terminate()`. It's now caught at the call site and prints `Error: <message>` + returns 1, matching the CLI's own clean-failure format exactly.

2. **Rounding/whitespace documented and tested, non-finite input rejected**: found the actual bug you were pointing at `std::stod("nan")` succeeds and returns a value that compares `false` against every relation (`<=`, `>`), so neither the existing `value <= 0.0` check nor the overflow check caught it, and the function went on to `static_cast<uint64_t>(NaN)`, which is undefined behavior (confirmed with a standalone repro before touching the real code: it silently produced `9223372036854775808`). Fixed with an explicit `std::isfinite` check. `inf`/`-inf` were already caught by the existing checks (verified this rather than assumed it), now covered by the same explicit check instead of incidentally. Rounding (round-half-up) and whitespace behavior (leading whitespace before the number is tolerated  that's `std::stod`'s own behavior anything after the number is not) are now documented on the function and covered by new tests.

**Validation**, run just now:

```
$ g++ -std=c++17 -I src -Wall -Wextra tests/cpp/test_cli_args.cpp src/cli/args.cpp -o test_cli_args
$ ./test_cli_args
All CLI parser tests passed
$ echo $?
0
```

```
$ g++ -std=c++17 -fsyntax-only -I src -I include examples/trtmc_dataset_benchmark.cpp
$ echo $?
0
```

I still don't have a CUDA/TensorRT toolchain here to produce a fully linked `trtmc_dataset_benchmark` binary, so the second command is a syntax/type check against the real project headers rather than an executed run of the benchmark itself  same limitation as the original PR, disclosed again here since the file changed since then.

chaofengw-nv:
Thanks for clarifying. Not having a local CUDA/TensorRT environment is fine. Your CPU-level tests and syntax check are helpful. we’ll complete the full CUDA/TensorRT build validation in our project CI before merging. 
If the full CUDA/TensorRT validation reveals any issues, I’d be happy to help investigate and contribute fixes directly to this PR. 

Separately, we’ve identified a permissions issue related to our recent CI changes that is temporarily preventing external contributors from triggering CI. This is on our side, and  we’re working to resolve it as quickly as possible.

lukiod:
Correction to my earlier note about not having a CUDA/TensorRT toolchain: that was true only because I hadn't actually checked for real GPU hardware or tried installing one — there is a real NVIDIA GPU on this machine, so I set up the actual toolchain (NVIDIA's own `nvcr.io/nvidia/tensorrt:26.07-py3` container image, the same one this repo's own `Dockerfile.dev.x86` uses) and did a full, real build and run, not another syntax check.

**Real build**: `cmake --build . --target trtmc_dataset_benchmark` — full link succeeded, including `trtmc_core` and every CUDA-touching translation unit.

**Real test suite**: `g++ -std=c++17 tests/cpp/test_cli_args.cpp src/cli/args.cpp` → `./test_cli_args` → `All CLI parser tests passed`, exit 0.

**Real CLI execution** (positional `<bundle> <dataset> <output>` args, `--kv-cache-size` as requested):

| Input | Result |
|---|---|
| `5GB` | parses; proceeds to (unrelated) dataset-load step |
| `1.5GB` | parses; proceeds to dataset-load step |
| `nan` | `Error: --kv-cache-size expects a positive size like 90GB or 90GiB`, exit 1 |
| `inf` | same rejection, exit 1 |
| `-inf` | same rejection, exit 1 |
| `" 5GB"` (leading space) | parses (matches `tests/cpp/test_cli_args.cpp`'s own documented `std::stod` leading-whitespace behavior) |
| `"5GB "` (trailing space) | rejected — trailing space becomes part of the unit suffix, fails exact match |
| `"5 GB"` (space between number and unit) | rejected, same reason |

All of this matches what's already documented and tested in the PR — this pass didn't change any code, it replaced an unverified disclaimer with an actual verified run.

yifeif-nv:
Hello! @lukiod  thanks for the contribution! We are experience some repo wide CI failures when migrating to the new community visible CI interface! And we are actively working on it. Will let you know once it's fixed

yifeif-nv:
Hey, looks like this PR has been stalled for a couple of days. Let me re-trigger the community CI for you, and please do let me know if we plan to push this PR through!

yifeif-nv:
It looks like this PR is not rebased onto the latest main top of the tree. Please go ahead and do so; that should trigger the CI for you.

Issue #976, opened 2026-08-21
[Refactor] Render E2E reproduction commands with shlex.join


chaofengw-nv:
## Parent

Inspired by #963: use an existing standard-library implementation rather than maintaining partial quoting logic.

## What to build

Make E2E reproduction-command providers return raw argv values consistently, then render a copy-pasteable shell command exactly once with `shlex.join` at the final presentation seam. Remove the harness's handwritten shell quote implementation and the model-owned pass-through wrappers that only duplicate `shlex.quote` behavior.

The current interface mixes raw tokens and pre-quoted tokens, and some final commands use a plain space join. Prompts containing spaces can therefore be rendered as different argv when copied back into a shell.

## Acceptance criteria

- [ ] Every reproduction-command provider returns raw argv tokens with no pre-quoting.
- [ ] Human-readable commands are rendered once with `shlex.join`.
- [ ] The handwritten harness quote implementation and shallow pass-through quote wrappers are removed.
- [ ] Round-trip tests cover spaces, empty tokens, single quotes, dollar signs, backslashes, and newlines.
- [ ] Existing model-owned command selection remains model-local and the rendered commands reconstruct the original argv exactly.

## Blocked by

None - can start immediately.

AbishekCoder1:
Hi @chaofengw-nv! I'd love to take this on as my first contribution. My plan is to strip out the manual string quoting logic in the providers, ensure they return raw argv lists, and implement shlex.join at the final presentation seam. I'll also add the requested round-trip tests for edge cases like spaces and newlines. Let me know if I'm good to start

yifeif-nv:
> Hi [@chaofengw-nv](https://github.com/chaofengw-nv)! I'd love to take this on as my first contribution. My plan is to strip out the manual string quoting logic in the providers, ensure they return raw argv lists, and implement shlex.join at the final presentation seam. I'll also add the requested round-trip tests for edge cases like spaces and newlines. Let me know if I'm good to start

Hi @AbishekCoder1, we've already have 2 duplicated PRs on this issue so this might not be a good issue for you to work on! But we are working on a task list that people can pick up please stay tuned for that and we will have a good list of tasks for people to work on soon ish

AbishekCoder1:
@yifeif-nv sure ping me once you have an issue to assign me

Issue #963, opened 2026-08-21
[Refactor] Replace custom JSON parser in cli_support.cpp with nlohmann::json


kanhaiya-dct:
### Request type

Improvement to existing behavior

## Description
While exploring the configuration parsing logic in `src/runtime/config/cli_support.cpp`, I noticed that the file implements its own custom "Tiny JSON parser" (the `Parser` struct) along with several custom manual JSON string serializers and string scanners (`match_object_end`, etc.). 
Since the project already includes and heavily relies on the robust `nlohmann/json` library in other core components (such as the tokenizers and plugins), maintaining a custom 300+ line JSON parser specifically for the CLI configuration is unnecessary. It increases technical debt, adds unnecessary binary size, and is prone to edge-case parsing bugs compared to a fully standard-compliant library.
## Motivation & Context
Replacing this custom implementation with `nlohmann::json` will:
1. **Reduce Code Complexity:** Eliminate over 300 lines of custom string manipulation and scanning logic.
2. **Increase Robustness:** Delegate JSON parsing to a well-tested, standard-compliant library.
3. **Unify Dependencies:** Ensure JSON parsing is handled consistently across the entire TensorRT-Model-Connect codebase.

## Proposed Solution
- Remove the custom `Parser` struct from `cli_support.cpp`.
- Remove manual JSON serialization functions (e.g., `append_json_escaped_string`, `try_append_numeric`).
- Refactor `parse_layered_json` and `bundle_to_effective_json` to utilize `nlohmann::json::parse()` and `.dump()`.
- Map the parsed JSON objects back to the internal `std::any` variants expected by the `ConfigBundle` schema.

### Submission checks

- [x] I searched open and closed issues and found no duplicate request.
- [x] I removed secrets, private/internal evidence, personal paths, and restricted artifacts.

yifeif-nv:
Looks good. Thanks for raising the PR and the contribution! I love those kinds of PRs that remove code rather than adding more.

yifeif-nv:
I'll leave this issue open until the remaining related issues are addressed

chaofengw-nv:
@kanhaiya-dct  Thanks for your contribution. Inspired by this, I have re‑checked some similar cases and filed relevant issues for tracking.

PR #1023, opened 2026-08-24
refactor(e2e): render repro commands once with shlex.join

dhruvdavest07:
## What

Repro command providers now return raw argv tokens, and `shlex.join` renders the shell-ready
string once at the point it is built. Removes the harness's own `_shell_quote` and the
thirteen per-model `_shell_quote` wrappers. Adds round-trip tests.

## Why

Quoting was happening in two different places and disagreeing about who owned it.

`tests/e2e_harness/orchestrator.py` carried a hand-written `_shell_quote`, and each of the
thirteen model repro plugins carried its own one-line wrapper around `shlex.quote`. Providers
pre-quoted some of their tokens, then the orchestrator joined the list with `" ".join`.

Two things fell out of that.

The obvious one: fifteen copies of logic the standard library already provides, and the
harness copy was a hand-rolled reimplementation rather than a call to `shlex.quote`.

The one that actually bites: the tokens nobody thought to pre-quote went into `" ".join`
bare. `ctx.engine_dir`, `ctx.binary_path`, `ctx.hf_python` and the bundle path are all
filesystem paths, none of them were quoted, and any one of them containing a space produced
a command that splits into the wrong argv when pasted into a shell. So
`--trtmc-binary /opt/my tools/trtmc` came out as two arguments.

Mixing the two conventions is what made it easy to miss. When some tokens arrive pre-quoted,
a bare token in the same list looks intentional.

## How

Providers hand back raw tokens. `shlex.join` is the single place a list becomes a string, so
there is one rule instead of a per-call judgement, and `shlex.quote` decides what needs
quoting.

`_shell_quote(str(x))` became `str(x)` at the three orchestrator call sites, and
`_shell_quote(x)` became `str(x)` in the plugins, keeping the coercion the wrapper was doing.
The four `" ".join(...)` calls that build `build_bundle`, `trt_inference`, `rerun_test` and
`rerun_test_rebuild` became `shlex.join(...)`.

`shlex.join` is available on every version this project supports (`requires-python = ">=3.10"`).

Two visible output changes, both intended:

- Paths containing spaces are now quoted, which is the fix.
- The `rerun_test` pytest node id comes out quoted, because
  `tests/test_e2e.py::test_e2e[case-name]` carries brackets that a shell treats as a glob. It
  was previously bare.

Everything that was already pre-quoted renders identically.

## Testing

New file `tests/e2e_harness/test_repro_commands.py`, 35 tests.

Round trip through `shlex.split` for prompt values covering spaces, the empty string, single
quotes, double quotes, dollar signs and `${}`, backslashes, newlines, tabs, `;`, `&&` and
`|`, glob characters, a trailing space, and a leading `#`.

Then the four cases the refactor actually changes: an engine dir with a space in
`build_bundle` and in `rerun_test`, a binary path with a space in `rerun_test` and in
`trt_inference`, an hf-python path with a space, and a bundle path with a space.

```
PYTHONPATH=python python -m pytest tests/e2e_harness/ -q
```

62 passed, which is the existing 27 plus these 35. `ruff check` is clean on every file
touched.

I ran the new tests against the previous implementation as a check on the tests themselves,
and it is worth reporting what that showed. My first pass at this file only exercised the
prompt values, and all of it passed against the old code too, because those tokens were
already being pre-quoted and the hand-rolled `_shell_quote` was, for those inputs, equivalent
to `shlex.quote`. Those tests were describing behaviour, not protecting it.

The six path cases are the ones that fail against the old implementation. The prompt cases
are kept as regression cover for the behaviour that was already correct, but the six are what
demonstrate the bug.

Unit level only. Running the E2E harness itself needs GPUs and built engines, which I do not
have, so this is verified through the argv the orchestrator renders rather than by executing
the commands.

## Notes

The diff spans fourteen files because the wrapper was copied into every model plugin, but each
plugin change is the same three lines: drop the import, drop the wrapper, call `str` instead.
Happy to split the harness change from the plugin cleanup if you would rather review them
separately.

Scope is limited to what #976 describes. #963's other items are untouched.

### DIFF
--- a/tests/e2e/models/deepseek_ocr/e2e_plugins/repro.py
+++ b/tests/e2e/models/deepseek_ocr/e2e_plugins/repro.py
@@ -5,13 +5,7 @@
 
 from __future__ import annotations
 
-import shlex
-
 from .contracts import E2ECase, ReproCommandProvider, RunContext
-
-
-def _shell_quote(value: object) -> str:
-    return shlex.quote(str(value))
 
 
 class DeepseekOcrReproCommandProvider:
@@ -36,9 +30,9 @@ class DeepseekOcrReproCommandProvider:
             "run",
             bundle_path,
             "--prompt",
-            _shell_quote(case.inputs.get("prompt", "Describe this image.")),
+            str(case.inputs.get("prompt", "Describe this image.")),
             "--image",
-            _shell_quote(image),
+            str(image),
             "--max-new-tokens",
             str(case.inputs.get("max_new_tokens", 30)),
         ]
--- a/tests/e2e/models/elf_flow/e2e_plugins/repro.py
+++ b/tests/e2e/models/elf_flow/e2e_plugins/repro.py
@@ -5,13 +5,7 @@
 
 from __future__ import annotations
 
-import shlex
-
 from .contracts import E2ECase, ReproCommandProvider, RunContext
-
-
-def _shell_quote(value: object) -> str:
-    return shlex.quote(str(value))
 
 
 class ElfFlowReproCommandProvider:
@@ -35,7 +29,7 @@ class ElfFlowReproCommandProvider:
             "run",
             bundle_path,
             "--prompt",
-            _shell_quote(
+            str(
                 case.inputs.get("prompt")
                 or case.inputs.get("source_text")
                 or case.inputs.get("condition_text")
--- a/tests/e2e/models/internvl/e2e_plugins/repro.py
+++ b/tests/e2e/models/internvl/e2e_plugins/repro.py
@@ -5,13 +5,7 @@
 
 from __future__ import annotations
 
-import shlex
-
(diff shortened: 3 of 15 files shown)


yifeif-nv:
Hey, @dhruvdavest07 thanks for contributing! There seems to be a minor issue on the community CI. Let me trigger that for you.!

yifeif-nv:
Thanks for the contribution. We currently have two overlapping implementations for #976, and we plan to proceed with this PR as the canonical one because it keeps quoting at the final rendering boundary and has the smaller, more focused diff. 

Please rebase onto the current main branch and fix the existing premerge failure. Once the updated head is green, we’ll review it for merge.

yifeif-nv:
@dhruvdavest07 Hello, just to follow up on this PR! can you help to rebase the PR onto the top of the tree in main so we can retrigger the internal CI for you?

Issue #955, opened 2026-08-20
Question: Qwen3 native KV-cache decode engine fails to build on SM120/SM121 (DGX Spark / GB10)


sparkleholic:
### Topic

Installation or source build

## Summary

On a DGX Spark (GB10, compute capability 12.1), `trtmc build` succeeds for Qwen2.5
models but fails for every Qwen3 model. The prefill engine always builds; the
**decode** engine fails inside TensorRT's Myelin attention fusion.

The failure is confined to the Qwen3 **native KV-cache** decode path
(`IAttention` + `key_value_lengths` + `IKVCacheUpdateLayer`). A correct
workaround exists (`--dynamic-kv-cache`, see below), so this is not a hard
blocker — but the current failure mode is an opaque internal TensorRT exception
rather than an actionable message, and the SM120/121 limitation is undocumented.

## Environment

| | |
|---|---|
| GPU | NVIDIA GB10 (DGX Spark), compute capability **12.1** (sm_121) |
| Driver | 580.173.02 |
| Arch / OS | aarch64, Ubuntu 24.04.4 LTS |
| CUDA | 13.0 (V13.0.88) |
| TensorRT | 11.1.0.106 (`libnvinfer-dev 11.1.0.106-1+cuda13.3`; pip `tensorrt==11.1.0.106`) |
| Build | host build (no container), `-DCMAKE_CUDA_ARCHITECTURES=121-real`, TRT backend ABI `11_1` |

The C++ build itself is fine: `trtmc`, `libtrtmc_backend_trt_11_1.so`, and
`libtrtmc_model_qwen.so` all compile, and Qwen2.5 bundles run correctly.

## Reproduction

```bash
trtmc build Qwen/Qwen3-0.6B -o qwen3-0.6b.bundle
```

```
[trtmc build] Prefill engine built [22.2s] (1144.9 MB)
Internal Error: MyelinCheckException: decompose_composite_ops.cpp:29:
  CHECK(composite_op->is_decomposable()) failed.
  Attention operation was not supported by a dedicated kernel.
[TRT] [E] Error Code: 10: Internal Error (Could not find any implementation for
  node {ForeignNode[(Unnamed Layer* 2) [Cast]...(Unnamed Layer* 2837) [Cast]]})
Error: dual-profile decoder engine build failed
```

## What we isolated

Every row below was run on this machine.

| Model | arch | head_dim | heads x head_dim vs hidden | QK-norm | Result |
[...]

PR #1028, opened 2026-08-25
fix(qwen): block unsafe SM121 native KV builds

JiaxinD:
## Background

Qwen3 native-KV engines on SM121 can build successfully with TensorRT ABI 11.1 or 11.2 and then return deterministic incorrect tokens, as reported in #955. A successful engine build is therefore not sufficient correctness evidence for this target/runtime tuple.

## Exit Criteria

- Reject affected dense Qwen3 native-KV builds before checkpoint weights are loaded.
- Keep Qwen2/Qwen2.5, SM120, SM103, qualified FP8 and other non-native routes, and explicitly identified future TensorRT ABIs outside the restriction.
- Fail closed on SM121 when the active CUDA device or TensorRT ABI cannot be identified.
- Document the exact user-visible boundary and provide an actionable diagnostic.

## Implementation

- Add a loader-safe, family-owned CUDA compute-capability probe and a narrow SM121/TensorRT ABI denylist to the Qwen native-KV routing contract.
- Invoke the platform validation from the existing pre-weight family hook. Direct plugin builds pass their resolved native-KV capability back into validation so qualified FP8 and tensor-parallel routes are not misclassified.
- Cover the unsafe/safe platform matrix, CUDA probe failures, plugin wiring, and non-native route preservation.
- Document the restriction in the README, Qwen quick start, and known issues.

## Validation

- `$env:PYTHONPATH = 'python;.'; py -3 -m pytest tests/e2e/models/qwen/test_qwen_native_kv_routing.py --confcutdir=tests/e2e/models/qwen --basetemp=.tmp/pytest-final -q`: 37 passed, 21 skipped because TensorRT is unavailable in the local Windows environment.
- `py -3 -m ruff check python/tensorrt_model_connect/families/qwen/build_routing.py python/tensorrt_model_connect/families/qwen/plugin.py tests/e2e/models/qwen/test_qwen_native_kv_routing.py`: passed.
- `$env:PYTHONPATH = 'python;.'; py -3 tools/model_ci.py validate`: passed for 84 families.
- `$env:PYTHONPATH = 'python;.'; py -3 tools/test_impact.py --validate`: passed with existing stale fallback-allowlist warnings.
- `py -3 tools/check_doc_file_references.py --strict website/docs`: passed with 0 errors and 0 warnings.
- `npm.cmd --prefix website exec -- docusaurus build website`: production site build passed.
- `git diff --check upstream/main...HEAD`: passed.
- `npm.cmd --prefix website run build`: the wrapper did not reach Docusaurus because its diagram check reports the unmodified `static/img/diagrams/architecture/build-ownership.svg` as stale on Windows.
- Community CPU / Required and all public child jobs passed for exact head `a2cb00fc6e1c58624284819a7e4b061aef12cc52`.

No SM121 GPU inference or parity run was available locally. Protected target-hardware validation remains required.

## Notes For Future Readers

Review the platform matrix and probe first, then the Qwen plugin hook, followed by the focused tests and documentation. The restriction is intentionally tied to confirmed ABI families rather than all SM12x devices or future TensorRT releases; an unknown ABI fails closed only on SM121.

This mitigates #955 by preventing silent wrong output; it does not claim that the underlying TensorRT kernel defect is fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,11 @@ trtmc run ./qwen3-0.6b.bundle --prompt "What is the capital of France? Answer in
 # Generated text: Paris
 ```
 
+Qwen3 native-KV builds are blocked on SM121 with TensorRT ABI 11.1 or 11.2
+because that combination can produce incorrect output. See
+[Known Issues](https://nvidia.github.io/TensorRT-Model-Connect/release-support/known-issues)
+for the exact boundary and alternatives.
+
 The same bundle works from
 [C++](https://nvidia.github.io/TensorRT-Model-Connect/api/cpp-api):
 
--- a/python/tensorrt_model_connect/families/qwen/build_routing.py
+++ b/python/tensorrt_model_connect/families/qwen/build_routing.py
@@ -10,6 +10,7 @@ import operator
 
 _INT32_MAX = (1 << 31) - 1
 _UINT64_MAX = (1 << 64) - 1
+_UNSAFE_NATIVE_KV_TRT_ABIS = {(12, 1): frozenset({"11.1", "11.2"})}
 
 
 class NativeKvCapability:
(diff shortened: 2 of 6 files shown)


JiaxinD:
@yifeif-nv This PR is ready for CI. Please trigger CI for the current head.

yifeif-nv:
Hi @JiaxinD Thanks for creating the fix. However, this fix will actually disable a bunch of platform support, and I think it is not the appropriate fix for this case. 

The bug is real. It is actually a TensorRT internal bug. I will be working on submitting a workaround PR to address this in the short term, basically by removing the internal masking inversion. New TRT version will contain the proper fix

Because of this, I'm proposing to close this PR. Would that be OK with you?

JiaxinD:
Sure, that works for me and thanks for letting me know! #1028 was intended as a temporary fail-closed guard while the SM121 failure was being investigated. Since the root cause is now confirmed as a TensorRT internal bug and #1035 provides the narrower workaround without dropping platform support, I will close #1028.

PR #1114, opened 2026-09-01
perf(moge): accelerate dynamic geometry path

yifeif-nv:
## Background

The original native MoGe fast path was fixed at 960x540, while the observed MoGe-2 workload spans portrait, landscape, and 4K inputs and is dominated by 1920x1080. Nsight showed that the remaining L4 gap came from decoder padding/layout traffic and full-resolution geometry materialization rather than builder optimization-level tuning.

## Exit Criteria

- Retain the original 960x540 contract and support the observed dynamic image range in the native TensorRT FP16 path, optimized for batch-1 1920x1080.
- Reduce the 1920x1080 GPU kernel sum below 35 ms on L4.
- Pass the unchanged seven geometry gates across the stored representative shapes and token grids, including 4K.
- Keep the implementation MoGe-owned and avoid ONNX, FP8, custom TensorRT plugins, runtime-core changes, and relaxed test criteria.

## Implementation

- Retain 960x540 while covering the observed 608x1074 through 3840x2160 shapes, with 1920x1080 as the optimization point.
- Keep transformer residual and decoder traffic in FP16 while preserving FP32 LayerNorm and the public FP32 geometry result.
- Replace each level-3 `HALF_PIXEL` resize, replicate pad, and 3x3 convolution with one mathematically equivalent native k6/s2 deconvolution.
- Use an explicit model-owned padding selection qualified against the complete representative-shape suite; all other convolutions retain replicate semantics.
- Emit compact engine-private affine-depth, validity, focal-sample, and metric-scale outputs instead of copying a full XYZ map and mask to the host.
- Recover focal length from finite interior samples whose mapped 3x3 validity neighborhood is complete, then reconstruct the unchanged public geometry result in the MoGe pipeline.
- Keep ordered RGB range checks so NaN, infinities, and out-of-range inputs still fail while avoiding redundant host classification work.
- Keep the broad FP32 attention graph decomposed with a family-owned builder optimization level of 0; the FP16 fast path remains at level 3.

### Change categories

- [ ] Public API
- [ ] ABI
- [x] Bundle or artifact contract
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

- `PYTHONPATH=python python3 -m pytest -q python/tensorrt_model_connect/families/moge/tests`: 34 passed on rebased head.
- `ruff check ...` and `ruff format --check ...` for the changed MoGe Python files: passed.
- `clang-format --dry-run --Werror src/runtime/models/moge/pipeline.cpp tests/cpp/models/moge/test_moge_pipeline.cpp`: passed.
- Focused native `test_moge_pipeline`: rebuilt and passed, including active engine-profile bounds, compact FP16 validity, invalid RGB values, edge/neighbor focal filtering, and public geometry reconstruction.
- `python -m tensorrt_model_connect build ... --method trt --precision fp16`: clean production graph built successfully with TensorRT 11.2.1.2; no quantization was enabled.
- Three independent Nsight runs of the established C++ SOL harness (10 warmups, five measured requests) on the final restored-profile plan reported 1920x1080 GPU kernel sums of `34.349`, `34.303`, and `35.241 ms/request`; median `34.349 ms`.
- The restored 960x540 FP16 path built and ran successfully and passed all seven unchanged geometry gates; intrinsics relative error was `0.000228` and mask IoU was `0.999676`.
- The clean production engine and clean C++ plugin passed all seven unchanged gates on 10/10 stored reference shapes, covering eight token grids and 3840x2160.

### Representative Accuracy Bounds

- Lowest mask IoU: `0.999156` (threshold `0.999`).
- Highest depth AbsRel: `0.003091` (threshold `0.005`).
- Highest depth relative L2: `0.014768` (threshold `0.02`).
- Highest points relative L2: `0.015077` (threshold `0.02`).
- Highest intrinsics relative error: `0.000967` (threshold `0.002`).
- All points cosine values exceeded `0.99999`; point/depth consistency was `0.0` for every shape.

- Current head: `37148e850865961b8bb6a3c4f4ad2ad68505b21d`, rebased on `github/main@591e67367a11629996cdcc1fb374f9b7e3baa7ad`.
- Hardware: NVIDIA L4, batch 1, 1920x1080 optimization shape, 1,825 runtime tokens.
- Container: `nvcr.io/nvidia/pytorch:26.08-py3@sha256:3becd068f49bd2ad38f90db5f9a4803019a76933a24e63d821376c44e7a9200a`; CUDA 13.4; TensorRT 11.2.1.2.
- Checkpoint: `Ruicheng/moge-2-vitl@39c4d5e957afe587e04eec59dc2bcc3be5ecd968`; SHA-256 `3eefd4abb2102f38f12b2d1992e5ff15e4923e5431c67dd494afe157e0111cd5`.
- Clean production bundle SHA-256: `551c2fad9e3f9f467ffdf444bdbf4efdf53e9fea8926eda1d2c314ebfb78e958`.

### Not Run / Remaining Gaps

- GitHub required checks passed on exact head `37148e850865961b8bb6a3c4f4ad2ad68505b21d`.
- `TRTMC Internal CI / Automated premerge gate` passed on the same exact head.

## Notes For Future Readers

- Review `families/moge/model.py` first for the dynamic mixed-precision graph and level-3 fusion, then `src/runtime/models/moge/pipeline.cpp` for the compact output and focal-recovery contract.
- The zero-padding selection is intentionally explicit and model-specific. Expanding it requires rerunning the full representative-shape geometry suite; global zero padding fails accuracy.
- The engine-private output contract changed, so bundles built from the earlier fixed-shape head must be rebuilt with the matching MoGe plugin. The public geometry API and ABI are unchanged.
- Packed attention, FP8, grouped paired heads, and additional non-exact linear folds were evaluated and rejected; none are present in this diff.

### Risk level

- [ ] Low
- [x] Medium
- [ ] High

Risk rationale: this changes the MoGe fast-path dynamic profile and engine-private output contract. The change is isolated to MoGe-owned files and is covered by exact-head native build, repeated L4 profiling, focused C++/Python tests, and 10-shape real-engine parity.


yifeif-nv:
Internal CI blocker isolated to the protected runner filesystem, not MoGe.

- Exact head/base: `2a5cd5cc2c381183910307cc06187d8ff9a5cf64` on `3b87adebdb4da59af00413b02f783df411620245`.
- The protected-style command (`python -m tools.ci pipeline premerge-unit`, scope `all`) reproduced on L4 `ipp2-2246` with TensorRT present: `1 failed, 3806 passed, 1 skipped`.
- Real failing node: `tests/tools/test_model_proof_runner.py::test_distinct_explicit_hf_cache_paths_reach_both_containers[premerge-bridge-False-False-True]`.
- Error: `cp --reflink=always` returned `Operation not supported` while materializing the selected HF cache. The private failure extractor appears to publish the pytest progress token as `Test: [100`, hiding this node ID.
- Negative control: unmodified `github/main@3b87adebdb4da59af00413b02f783df411620245` fails both parameterizations of the same test on that filesystem with the same reflink error.
- Community CI explicitly deselects this test because its scratch filesystem is not guaranteed to support reflinks; protected scope intentionally retains it.

Please provision a reflink-capable protected scratch/TMP filesystem (or otherwise repair the protected runner contract), then rerun the exact head. I did not change the test, its passing criteria, or the MoGe implementation to bypass this gate.

PR #1144, opened 2026-09-03
test(phi4-multimodal): cover runtime JSON contract

Moviw:
## Background

Migrate this test PR to the family-owned architecture requested in discussion #1183. The current consumer reads explicit preprocessing fields from `runtime.json`; the retired `parse_base_config()` and `MODEL.toml` registration no longer describe this boundary.

## Exit Criteria

- Exercise `phi4_multimodal_parse_preprocess_config()`, the production plugin's preprocessing consumer, through a CPU CTest.
- Pin top-level field binding, template escapes, required types, geometry and interpolation validation.
- Keep all changes inside `families/phi4_multimodal/`.

## Implementation

Add a family-local C++ test and register it in `runtime/CMakeLists.txt` with the `cpu` label. The executable compiles `image_preprocessor.cpp` directly and uses the existing nlohmann/STB dependencies, without linking the family DSO.

The fixture places nested decoys before explicit top-level preprocessing fields and uses sentinel values to expose ignored inputs. Checks remain active in Release builds. Missing/null fields, non-integer dimensions, malformed triplets, non-positive geometry, unsupported interpolation and invalid JSON are rejected; all three supported interpolation modes are preserved.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [x] CI or developer tooling

### Commands and Results

Direct source-only build and run passed (set `JSON_INCLUDE` to the installed nlohmann include directory):

```sh
g++ -std=c++17 -O2 -I. -Ithird_party/stb -I"$JSON_INCLUDE" \
  families/phi4_multimodal/tests/cpp/test_phi4_multimodal_runtime_config_contract.cpp \
  families/phi4_multimodal/runtime/image_preprocessor.cpp \
  -o /tmp/test_phi4_multimodal_runtime_config_contract
/tmp/test_phi4_multimodal_runtime_config_contract
# PASS: Phi4 runtime preprocessing config contract
```

- Family-only CMake harness loading the actual `runtime/CMakeLists.txt`, followed by `cmake --build <build> --target test_phi4_multimodal_runtime_config_contract` and `ctest --test-dir <build> -R '^test_phi4_multimodal_runtime_config_contract$' --output-on-failure --no-tests=error`: 1/1 passed, CPU-labelled. The harness supplies the repository root and existing nlohmann include target; it does not configure the full project or use SDK stubs.
- Three independent production-source mutations (hardcode vision width, remove geometry guard, remove interpolation guard): each compiled and made the test exit 1. Mutations were confined to disposable copies.
- `python3 -m tools.model_ci validate`: passed.
- `python3 tools/test_impact.py --validate`: passed.
- `python3 -m pytest tools/tests/test_architecture.py tools/tests/test_family_impact.py tools/tests/test_community_ci.py tools/tests/test_public_source_hygiene.py tools/tests/test_new_ci.py tools/tests/test_pr_metadata.py -q -p no:cacheprovider`: 125 passed.
- `clang-format --dry-run --Werror families/phi4_multimodal/tests/cpp/test_phi4_multimodal_runtime_config_contract.cpp` and `git diff --check`: passed.

- Base: `40523401ae65735056f85ae73028c04025aa1c50`.
- Head: `4be682035a0af4582a938402b3c3aa6e0e99ab6e`.
- Local Linux x86_64, GCC 11.4, C++17, Release build. Inline synthetic JSON; no checkpoint or GPU needed for this test.

### Not Run / Remaining Gaps

Full root configure (`cmake -S . -B <build> -G Ninja -DTRTMC_BUILD_TESTS=ON -DTRTMC_ENABLE_BYOK=OFF`) stopped because `NvInferRuntime.h` is unavailable locally. Full project integration and protected premerge remain for exact-head CI. Decoder loading, image processing, audio and inference are outside this preprocessing-config unit.

## Notes For Future Readers

This replaces the pre-cutover implementation and its validation record. Review the family CMake registration, then the fixture and rejection checks. Production parser behavior is unchanged; keep runtime.json coverage here rather than restoring old checkpoint-config fallback expectations.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High

Test-only, one family; existing production and GPU targets are unchanged.

### DIFF
--- a/families/phi4_multimodal/runtime/CMakeLists.txt
+++ b/families/phi4_multimodal/runtime/CMakeLists.txt
@@ -44,3 +44,26 @@ set_target_properties(trtmc_model_phi4_multimodal PROPERTIES
 install(TARGETS trtmc_model_phi4_multimodal
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
 )
+
+if(TRTMC_BUILD_TESTS)
+  add_executable(test_phi4_multimodal_runtime_config_contract
+    ${PROJECT_SOURCE_DIR}/families/phi4_multimodal/tests/cpp/test_phi4_multimodal_runtime_config_contract.cpp
+    image_preprocessor.cpp
+  )
+  target_include_directories(test_phi4_multimodal_runtime_config_contract PRIVATE
+    ${PROJECT_SOURCE_DIR}
+  )
+  target_include_directories(test_phi4_multimodal_runtime_config_contract SYSTEM PRIVATE
+    ${PROJECT_SOURCE_DIR}/third_party/stb
+  )
+  target_link_libraries(test_phi4_multimodal_runtime_config_contract PRIVATE
+    nlohmann_json::nlohmann_json
+  )
+  target_compile_options(test_phi4_multimodal_runtime_config_contract PRIVATE
+    -Wall -Wextra -Wpedantic
+  )
+  add_test(NAME test_phi4_multimodal_runtime_config_contract
+    COMMAND test_phi4_multimodal_runtime_config_contract
+  )
+  set_tests_properties(test_phi4_multimodal_runtime_config_contract PROPERTIES LABELS cpu)
+endif()
--- a/families/phi4_multimodal/tests/cpp/test_phi4_multimodal_runtime_config_contract.cpp
+++ b/families/phi4_multimodal/tests/cpp/test_phi4_multimodal_runtime_config_contract.cpp
@@ -1,0 +1,125 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "families/phi4_multimodal/runtime/image_preprocessor.h"
+
+#include <cmath>
+#include <iostream>
+#include <nlohmann/json.hpp>
+#include <stdexcept>
+#include <string>
+
+namespace {
+int failures = 0;
+
+void check(bool condition, const std::string& message) {
+    if (!condition) {
+        std::cerr << "FAIL: " << message << '\n';
+        ++failures;
+    }
+}
+
+void rejects(const std::string& text, const std::string& message) {
+    try {
+        (void)trtmc::phi4_multimodal_parse_preprocess_config(text);
+        check(false, message);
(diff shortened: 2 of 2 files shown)


Moviw:
Addressed — `image_token_id` and `vision_output_dim` are now asserted, with strict non-default fixture values, through `phi4_multimodal_parse_preprocess_config()`: the same function `plugin.cpp` calls to build its `Phi4MultimodalPreprocessConfig`. Asserting them via `extract_json_int` would only have tested the shared JSON helper, not this family's consumption (same point you raised on #1143).

`img_processor` now carries decoy `image_token_id`, `vision_output_dim` and `num_image_pad_tokens` values, so a consumer reaching into the vision tower fails. Removing the promoted top-level keys fails the VL half:

```
FAIL: image token id reaches the VL config
FAIL: vision output dim is the decoder width
FAIL: image pad token count
```

`has_vision_engine` is the one field left unasserted. The plugin reads it inline while assembling its `PipelineContext` to decide whether to load the vision module, so reaching that consumer needs a bundle with engines — it stays with the `REQUIRES_TRT,REQUIRES_GPU` pipeline test rather than being restated here as a generic JSON lookup. The PR description now says so explicitly instead of claiming coverage for all three.

Moviw:
@yifeif-nv This PR is ready for CI. Please trigger CI for the current head (`6d74793e`).

Community CPU is green on this head. The CodeRabbit comment about covering `image_token_id` and `vision_output_dim` is addressed: both are now asserted through `phi4_multimodal_parse_preprocess_config()`, the function `plugin.cpp` itself calls, with decoy values in `img_processor` so a consumer reaching into the vision tower fails. `has_vision_engine` is the one field left uncovered and the description says why.

yifeif-nv:
Hey @Moviw Thanks for contributing! PR looks good and I'll trigger the internal CI for you

Moviw:
Updated this PR for the architecture in #1183. The test and CMake registration now live under families/phi4_multimodal/ and cover the current runtime.json preprocessing contract. Local CPU tests and mutation checks pass; the PR description contains the updated validation details. Ready for another look!

yifeif-nv:
> Updated this PR for the architecture in #1183. The test and CMake registration now live under families/phi4_multimodal/ and cover the current runtime.json preprocessing contract. Local CPU tests and mutation checks pass; the PR description contains the updated validation details. Ready for another look!

Thanks for the migration! Triggering internal CI now

PR #1150, opened 2026-09-03
feat(timm_regnet): add timm RegNet image-classification family

zhenshanx-nv:
## Background

RegNet is the last of the tensorrtx classifier baselines that reuses the
existing convolutional op set. `timm/regnety_040.ra3_in1k` cannot be built or
served today.

## Exit Criteria

- A `timm_regnet` family builds timm RegNet checkpoints from HF-hosted
- RegNetX (no squeeze-excite) and RegNetY (with it) build from one code path.

Non-goals: quantized builds, tensor-parallel builds, and the RegNetZ variants,
which change the block structure.

## Implementation

The layout is recovered from the checkpoint: stages and block counts from the
`s<stage>.b<block>` keys, the convolution group count from the 3x3 weight shape,
and the squeeze-excite gate and downsample path from the keys that are present.
RegNet halves the resolution in the first block of every stage, so the stride is
a uniform rule rather than a per-model table.

Detecting the gate from the keys means RegNetX builds from the same path as
RegNetY with no extra branch.

Two shape details worth noting for reviewers:

- timm wraps each convolution and its norm in a `ConvNormAct`, so weights are
  named `conv.weight` and `bn.*` under each leaf rather than flat `convN`/`bnN`.
- The squeeze-excite projections are named `fc1`/`fc2` even though both are 1x1
  convolutions.

The gate uses a ReLU inner activation with a plain sigmoid. That is a third
combination, distinct from MobileNetV3 (hard-sigmoid, ReLU) and EfficientNet
(sigmoid, SiLU), so the family keeps its own copy rather than sharing a
parameterised helper.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 3990 passed, 8 skipped

python -m pytest -q tests/e2e/models/timm_regnet/test_timm_regnet_family_plugin.py
=> 15 passed

cmake --build $BUILD --target trtmc_model_timm_regnet \
  test_timm_regnet_image_preprocess_seam
$BUILD/test_timm_regnet_image_preprocess_seam

clang-format => clean
```

Numerical parity against timm's own implementation, which shares no code with
the builder:

| --- | --- | --- | --- |
| `timm/regnety_040.ra3_in1k` | 0.99999874 | match | 5/5 |

- `timm/regnety_040.ra3_in1k` @ `a1875159c7b04f3b3189af804642eebba1e2c118`.

### Not Run / Remaining Gaps

- Only `regnety_040` was verified numerically. RegNetX is handled by the same
  path and is unit-tested for the no-gate case, but **no RegNetX checkpoint was
  downloaded or compared**, so that path has no numerical evidence.
- The `regnetz_*` variants match the `regnet` prefix but change the block
  structure and are not supported.

## Notes For Future Readers

The three squeeze-excite variants across these families differ in both the inner
activation and the gate, and none of the differences changes tensor shapes. If
you add another family with a gate, check both against timm rather than assuming
one of the existing copies applies.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1207, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1152, opened 2026-09-03
feat(timm_ghostnet): add timm GhostNet image-classification family

zhenshanx-nv:
## Background

GhostNet is one of the remaining efficient-classifier baselines in the tensorrtx
set. `timm/ghostnet_100.in1k` cannot be built or served today.

## Exit Criteria

- A `timm_ghostnet` family builds timm GhostNet checkpoints from HF-hosted

Non-goals: quantized builds, tensor-parallel builds, and GhostNetV2, which adds
a decoupled attention branch this family does not build.

## Implementation

A Ghost module produces half its channels with a pointwise convolution and the
other half with a cheap depthwise convolution applied to that result, then
concatenates the two. Within a bottleneck the first Ghost module activates and
the second does not.

The whole layout is recovered from the checkpoint, **including the stride**: a
bottleneck downsamples exactly when it carries a `conv_dw` depthwise convolution
between its two Ghost modules. Together with DenseNet this is one of only two
families here that needs no architecture table at all. The squeeze-excite gate
and the four-layer projection shortcut (depthwise, norm, pointwise, norm) are
likewise detected from the keys.

The gate uses a ReLU inner activation with a hard-sigmoid, matching MobileNetV3
rather than the EfficientNet or RegNet combinations.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 3990 passed, 8 skipped

python -m pytest -q tests/e2e/models/timm_ghostnet/test_timm_ghostnet_family_plugin.py
=> 14 passed

cmake --build $BUILD --target trtmc_model_timm_ghostnet \
  test_timm_ghostnet_image_preprocess_seam
$BUILD/test_timm_ghostnet_image_preprocess_seam

clang-format => clean
```

Numerical parity against timm's own implementation, which shares no code with
the builder:

| --- | --- | --- | --- |
| `timm/ghostnet_100.in1k` | 0.99999901 | match | 5/5 |

- `timm/ghostnet_100.in1k` @ `e524c2ee5d5a9412f802ce38ff895aba07b30910`.

### Not Run / Remaining Gaps

- Only `ghostnet_100` was verified numerically. The other widths share the
  structure, which is fully derived, so they are expected to work, but none was
  downloaded.
- `ghostnetv2_*` matches the `ghostnet` prefix but adds a decoupled attention
  branch that this builder does not implement. Unlike the MNASNet gate case
  there is no explicit rejection for it, because its extra keys would fail the
  block classification and raise there instead.

## Notes For Future Readers

The cheap depthwise branch runs on the **output of the primary convolution**,
not on the block input. Wiring it to the input keeps every shape valid and only
changes the numbers, so verify against timm rather than by inspection.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1210, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1154, opened 2026-09-03
feat(timm_seresnet): add timm SE-ResNet image-classification family

zhenshanx-nv:
## Background

`timm_resnet` (#1121) covers the gate-free ResNet and ResNeXt variants. The
squeeze-excite variants are a separate timm architecture family and are not
supported: `timm/seresnet50.a1_in1k` cannot be built or served today.

## Exit Criteria

- A `timm_seresnet` family builds timm SE-ResNet checkpoints from HF-hosted
- The family's prefixes are disjoint from `timm_resnet`, so no checkpoint is
  claimed by both.

Non-goals: quantized builds and tensor-parallel builds.

## Implementation

The stage and block layout is recovered from the checkpoint exactly as in
`timm_resnet`: block counts from the `layerN.M` indices, basic versus bottleneck
from the presence of `conv3`, and the group count from the `conv2` input-channel
ratio.

Each block carries a squeeze-excite gate applied to its output **after the final
batch norm and before the residual add**. The gate uses a ReLU inner activation
with a plain sigmoid; timm names its two 1x1 projections `fc1` and `fc2`.

The gate is **required**, not optional. A checkpoint without one is rejected
rather than built as a plain ResNet, which would produce correct shapes and
wrong numbers. `timm_resnet` matches `resnet`/`resnext`/`wide_resnet` while this
family matches `seresnet`/`seresnext`/`legacy_seresnet`, so the two never
compete for the same checkpoint.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 3990 passed, 8 skipped

python -m pytest -q tests/e2e/models/timm_seresnet/test_timm_seresnet_family_plugin.py
=> 14 passed

cmake --build $BUILD --target trtmc_model_timm_seresnet \
  test_timm_seresnet_image_preprocess_seam
$BUILD/test_timm_seresnet_image_preprocess_seam

clang-format => clean
```

| --- | --- | --- | --- |
| `timm/seresnet50.a1_in1k` | 0.99998800 | match | 5/5 |

- `timm/seresnet50.a1_in1k` @ `23f3482ee9acc4f51a2668a7cdf255fbd9420417`.

### Not Run / Remaining Gaps

- Only `seresnet50` was verified numerically. The SE-ResNeXt grouped variants
  reuse the group derivation already exercised by `timm_resnet`, and the deeper
  SE-ResNets reuse the layout discovery, but **neither was downloaded or
  compared**.
- `legacy_seresnet*` matches the prefixes but has a different stem and gate
  placement; it is not verified and would likely need its own handling.

## Notes For Future Readers

This is the fourth family here with a squeeze-excite gate, and the combinations
still differ: MobileNetV3 uses ReLU with hard-sigmoid, EfficientNet SiLU with
sigmoid, RegNet and this family ReLU with sigmoid. None of the differences
changes a tensor shape, so check against timm rather than assuming.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1208, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1156, opened 2026-09-03
feat(timm_xception): add timm Xception image-classification family

zhenshanx-nv:
## Background

Xception is one of the remaining classifier baselines in the tensorrtx set.
`timm/xception41.tf_in1k` cannot be built or served today.

## Exit Criteria

- A `timm_xception` family builds timm's aligned Xception checkpoints from
  HF-hosted safetensors and produces logits matching timm's own implementation.

Non-goals: quantized builds, tensor-parallel builds, and the original
(non-aligned) Xception, which has a different module layout.

## Implementation

Most of the layout is recovered from the checkpoint: the block count from the
`blocks.<n>` keys, and the stride from whether a block carries a projection
shortcut, since Xception downsamples exactly in the blocks that project.

One thing is **not** recoverable, because activations carry no weights: the
final block is built differently from the rest.

| | blocks 0..n-1 | exit block |
| --- | --- | --- |
| leading ReLU before each separable conv | yes | no |
| ReLU inside the separable conv | no | yes, after both norms |
| residual add | yes | no |

That is keyed on the block being last. It was read out of timm's module
construction rather than guessed, and the numerical check confirms it.

Two further details, both confirmed by querying timm: the batch-norm epsilon is
`1e-3` (TensorFlow), and it is a 299x299 model normalised to `[-1, 1]`.

The family deliberately has **no input-divisibility guard**. The strided
convolutions pad by one, so 299 halves cleanly to 150, 75, 38 and so on. The
guard copied from the ResNet families rejected the model's own default input
size, which is how the omission was found.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 3990 passed, 8 skipped

python -m pytest -q tests/e2e/models/timm_xception/test_timm_xception_family_plugin.py
=> 15 passed

cmake --build $BUILD --target trtmc_model_timm_xception \
  test_timm_xception_image_preprocess_seam
$BUILD/test_timm_xception_image_preprocess_seam

clang-format => clean
```

| --- | --- | --- | --- |
| `timm/xception41.tf_in1k` | 0.99999929 | match | 5/5 |

- `timm/xception41.tf_in1k` @ `8a17189361e63c972815ef62f2a30dd5b9f393b1`.

### Not Run / Remaining Gaps

- Only `xception41` was verified numerically. `xception65` and `xception71`
  differ only in block count, which is derived, so they are expected to work,
  but neither was downloaded.
- The original non-aligned `xception` is matched by the `xception` prefix but
  has a different module layout. Its checkpoint would fail the block scan rather
  than build incorrectly, but this is not covered by a test.

## Notes For Future Readers

The exit block's inverted activation placement is the trap in this family.
Building it like the others keeps every tensor shape valid and changes only the
numbers, so verify against timm rather than by reading the checkpoint.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1206, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1159, opened 2026-09-04
feat(devtoolkit): add Docker target lifecycle

chaofengw-nv:
## Background

DevToolkit could bind and attest an already-running Docker container, but it had no composable capability to pull or build an image and create or start that container. Callers therefore had to implement Docker lifecycle outside the toolkit before environment resolution could begin.

## Exit Criteria

- Let callers independently resolve, provision, and attest an execution target before resolving CUDA/TensorRT.
- Support caller-selected Docker image references or Dockerfiles, GPUs, bind mounts, workspace/state paths, environment, IPC, and shared memory.
- Make retries idempotent and fail closed on name collisions, configuration drift, daemon drift, inspect failures, or undeclared effective configuration.
- Keep the capability independent from workflows, campaigns, cohorts, model-family semantics, and any one target provider's lifecycle policy.

## Implementation

- Adds `toolkit.targets.resolve()`, `provision()`, `ensure()`, and `attest()` plus a public `TargetProvider` extension contract.
- Separates provider-neutral target contracts and orchestration from the Docker adapter. The generic service forwards opaque provider-owned policy objects; `DockerTargetPolicy` defines only Docker lifecycle semantics.
- Moves built-in adapter wiring into a composition root and consolidates shared Docker CLI/inspect/daemon mechanics in a deep support module.
- Adds typed Docker image, GPU, mount, and target requests with explicit `ADOPT`, `START`, `ENSURE`, and `CREATE` policies.
- Pulls or builds images as requested, creates and starts containers using immutable image IDs, and returns an `ExecutionTarget` for the existing environment API.
- Records deeply immutable target plans and attested receipts while hashing, rather than serializing, environment and build-argument values.
- Matches the complete effective environment and mount set while allowing image defaults and image-declared volumes. Undeclared additions fail closed.
- Never deletes or silently replaces an existing container; mismatches report the fields that drifted.
- Resolves an omitted Docker command to one effective default used by plan identity, container creation, and existing-container matching.
- Updates the sample to compose target lifecycle with arbitrary-version environment resolution and TRTMC build/run capabilities.

### Change categories

- [x] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [x] CI or developer tooling

### Commands and Results

- `python3 -m pytest tests/tools/test_devtoolkit_capabilities.py tests/tools/test_devtoolkit_targets.py -q`: 94 passed on `413dd4a9a8e816e4f1ca7018235eb62e7e7aad0f`.
- `pre-commit run --files scripts/devToolkit/trtmc_devtoolkit/docker_target.py tests/tools/test_devtoolkit_targets.py`: all applicable hooks passed on the current head.
- Manual local Docker exact-head smoke with `command=None`, an existing `alpine:latest` image, and `PullPolicy.NEVER`: returned `created`, then `adopted`, created `Cmd=["sleep", "infinity"]`, and rejected a same-image container with `Cmd=["tail", "-f", "/dev/null"]`.
- Manual GPU-backed smoke on P2021 GPU 0 at parent `2c79710e26b0042b4e47eae8179de228bdc34028`: returned `created`, then `adopted`, with an unchanged target ID; the container saw `NVIDIA DRIVE P2021`.
- Manual GPU-backed smoke on GB300-1 GPU 1 at parent `2c79710e26b0042b4e47eae8179de228bdc34028`: returned `created`, then `adopted`, with an unchanged target ID; the container saw `NVIDIA GB300` and TensorRT `11.2.0.113`.
- All dedicated smoke-test containers and both remote staging directories were removed after validation.

- CPU tests and local real-Docker smoke tested head: `413dd4a9a8e816e4f1ca7018235eb62e7e7aad0f`.
- GPU-backed P2021 and GB300-1 smoke tested parent: `2c79710e26b0042b4e47eae8179de228bdc34028`.
- Local x86_64 Linux host with Docker client/server 27.4.1 and an existing CPU image.
- P2021 GPU 0 and GB300-1 GPU 1 using pre-existing local images with `PullPolicy.NEVER`; GB300-1 additionally verified TensorRT `11.2.0.113` inside the target.

### Not Run / Remaining Gaps

- The Dockerfile build path is covered by contract tests but not a real Docker build in this validation pass.
- The broader `tests/tools` collection was not counted because this shell resolves an unrelated editable benchmark package from another worktree, causing four pre-existing import errors during collection. The directly affected DevToolkit suites run in isolation and pass.
- GitHub checks for the new head are pending and are not represented as passing here.

## Notes For Future Readers

- Review `target_contracts.py` and `target_service.py` first for the provider-neutral seam, `docker_target.py` for Docker lifecycle semantics, `docker_support.py` for shared Docker transport mechanics, and `builtin_registry.py` for composition.
- Target lifecycle and environment/toolchain provisioning are deliberately separate. A provisioned target feeds its immutable `execution_target` into `EnvironmentRequest`; no workflow DAG is introduced.
- Provider-specific lifecycle behavior belongs in each adapter. Adding an SSH, Kubernetes, or other target must not add its request types or policy semantics to the generic target service.
- No legacy plan/apply compatibility layer is added. Existing direct adoption through `ExecutionTarget.docker()` remains a lower-level capability.

### Risk level

- [ ] Low
- [x] Medium
- [ ] High

This adds a public API and controlled Docker mutations. Risk is bounded by explicit adapter-owned policies, immutable image/container binding, plan ownership labels, fail-closed inspection and drift checks, no automatic deletion/replacement, contract coverage, and real Docker lifecycle smoke tests.


sylvesterkaczmarek (commented):
One target-identity gap: `command=None` creates the container with the effective command `sleep infinity`, but `_mismatches()` skips command comparison whenever the request value is `None`. `ENSURE` can therefore adopt an existing same-image container running an arbitrary command even though a newly created target would not. Could the comparison use the effective default command as well?

chaofengw-nv:
@sylvesterkaczmarek Thanks for catching this. Fixed in `413dd4a9a8e816e4f1ca7018235eb62e7e7aad0f`, with a regression test added.

PR #1165, opened 2026-09-04
feat(timm_mobilenetv2): add timm MobileNetV2 image-classification family

zhenshanx-nv:
## Background

MobileNetV2 is one of the remaining mobile classifier baselines in the tensorrtx
set. `timm/mobilenetv2_100.ra_in1k` cannot be built or served today.

## Exit Criteria

- A `timm_mobilenetv2` family builds timm MobileNetV2 checkpoints from HF-hosted

Non-goals: quantized builds and tensor-parallel builds.

## Implementation

Block shape is recovered from the checkpoint: the block kind follows from which
convolutions are present and the depthwise kernel from its weight shape. The
activation is uniform ReLU6, so only the per-stage stride comes from an
architecture table.

MobileNetV2 carries no squeeze-excite gate. The gate is still detected from the
keys and a checkpoint that has one is **rejected** rather than built with the
gate silently dropped, which would give correct shapes and wrong numbers.

The only new op is ReLU6, expressed as a clamp to `[0, 6]`.

Structurally this family is close to `timm_mnasnet`: the same key names, the
same head order, and the same 7-stage stride schedule, differing only in the
activation. Per the repository's model-family ownership rule the code is
duplicated rather than shared.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 4042 passed, 8 skipped

python -m pytest -q \
  tests/e2e/models/timm_mobilenetv2/test_timm_mobilenetv2_family_plugin.py
=> 15 passed

cmake --build $B --target trtmc_model_timm_mobilenetv2 \
  test_timm_mobilenetv2_image_preprocess_seam
$B/test_timm_mobilenetv2_image_preprocess_seam

clang-format => clean
```

| --- | --- | --- | --- |
| `timm/mobilenetv2_100.ra_in1k` | 0.99997847 | match | 5/5 |

The first parity attempt reported a failure that turned out to be a defect in
the **test harness**, not the model: the verification script had been derived
from the Inception one and still fed a 299x299 input, which timm accepts through
its adaptive pooling while the engine builds the 224x224 input its config
declares. Reimplementing the block order in plain PyTorch matched timm exactly
(correlation 1.0, maximum difference 0.0), which localised the fault to the
harness. Every other family's script was audited afterwards and feeds the size
its own checkpoint declares.

- `timm/mobilenetv2_100.ra_in1k` @ `5afa12513b048c79b9147a5d210e9bdc50035481`.

### Not Run / Remaining Gaps

- Only the `_100` width was verified numerically. Other widths share the
  schedule and are expected to work, but none was downloaded.

## Notes For Future Readers

When adding a family by adapting another one's verification script, check the
input size first. timm's adaptive pooling accepts a wrong size silently, so the
comparison fails in a way that looks like a model defect.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Closing as redundant: the timm_mobilenetv2 family landed in #1178, so this adds nothing. It also targets the pre-cutover layout removed by #1188.

PR #1166, opened 2026-09-04
feat(timm_inception_v4): add timm Inception-v4 image-classification family

zhenshanx-nv:
## Background

`timm_inception` (#1155) covers Inception-v3. Inception-v4 is a different
architecture with its own block set and is not supported:
`timm/inception_v4.tf_in1k` cannot be built or served today.

## Exit Criteria

- A `timm_inception_v4` family builds timm Inception-v4 checkpoints from
  HF-hosted safetensors and produces logits matching timm's own implementation.
- The family's prefixes are disjoint from `timm_inception`.

Non-goals: quantized builds, tensor-parallel builds, and Inception-ResNet-v2,
which adds residual scaling.

## Implementation

Inception-v4 is a flat `features` sequence of 22 blocks in eight shapes. Each
block is classified from the keys the checkpoint carries for it, so the block
order is read off the checkpoint and only the branch wiring is written out.

Two topologies cannot be separated by their top-level branch names, because the
pooling branches carry no weights:

- **Reduction-A** is recognised structurally: its first branch is a single
  convolution where the others are chains.
- **Mixed4a and Reduction-B cannot be separated by shape at all.** Their
  branches are identical and they differ only in stride, which the checkpoint
  does not record. Ordering is therefore the only honest discriminator: the
  first such block is Mixed4a, any later one is Reduction-B. This is stated
  plainly in the code rather than hidden behind a magic index.

The average pool in the pooled branches **excludes** the zero padding, the
opposite of Inception-v3. Both were read out of timm rather than assumed, and
v3's use of the including form was rechecked while adding this family, since
getting it backwards changes only border values and no shapes.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
=> 4042 passed, 8 skipped

python -m pytest -q \
  tests/e2e/models/timm_inception_v4/test_timm_inception_v4_family_plugin.py
=> 17 passed

cmake --build $B --target trtmc_model_timm_inception_v4 \
  test_timm_inception_v4_image_preprocess_seam
$B/test_timm_inception_v4_image_preprocess_seam

clang-format => clean
```

| --- | --- | --- | --- |
| `timm/inception_v4.tf_in1k` | 0.99999951 | match | 5/5 |

The state dict loads into timm with no missing or unexpected keys. Each block
topology is also unit-tested for correct identification, including the two cases
that share branch names.

- `timm/inception_v4.tf_in1k` @ `ad5d294cda312745a9afa433e45d9ccd956a1548`.

### Not Run / Remaining Gaps

- The Mixed4a versus Reduction-B ordering rule holds for the published
  Inception-v4 topology. A variant that placed two Mixed4a-shaped blocks before
  the first reduction would be mis-built, and nothing in the checkpoint would
  reveal it. No such variant is known, but the assumption is worth stating.
- Only the `tf_in1k` weights were verified.

## Notes For Future Readers

The pooled branches differ between v3 and v4 in whether the average counts the
zero padding. That difference does not change any tensor shape, so it will not
surface as an error, only as slightly wrong numbers at the borders.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1212, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1167, opened 2026-09-05
feat(timm_inception_resnet): add timm Inception-ResNet image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no Inception-ResNet family. Inception-ResNet-v2 is one
of the classifiers listed by wang-xinyu/tensorrtx, and it is the last Inception
variant we do not cover after Inception-v3 (#1155) and Inception-v4 (#1166). It
is the first family here that mixes Inception-style concatenated branches with
scaled residual additions, so it does not fit either existing plugin.

## Exit Criteria

- `timm/inception_resnet_v2.tf_in1k` builds and classifies through the
  `timm_inception_resnet_image_classification` runtime strategy.

`NotImplementedError`; no other Inception-ResNet checkpoint is registered.

## Implementation

New `timm_inception_resnet` family, following the layout the other timm families
use: a Python builder plugin that emits the network through the TensorRT Network
API, a C++ runtime model with its own image-preprocess seam, and an e2e model
directory with one manifest.

The depth is read from the checkpoint rather than tabulated: `_discover_layout`
counts the `repeat`, `repeat_1`, and `repeat_2` block indices and rejects
non-contiguous ones. Two things the weights do not record are stated as
constants with a comment saying why: the per-group residual scale
(0.17 / 0.10 / 0.20) and the fact that the trailing `block8` adds unscaled and
omits the activation.

Two details differ from the sibling Inception families and are easy to get
wrong:
- Batch norm uses eps 1e-3, the TensorFlow value, because this is a TF port.
- The 1x1 projection inside each residual block has a bias and no batch norm,
  unlike every other convolution in the model.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 262.21s
```

The 3 failures are `tests/tools/test_model_proof_runner.py` cases that call
`/bin/cp --reflink` on a filesystem that does not support cloning. They fail the
same way on an unmodified checkout of this machine and are unrelated to this
change.

random 1x3x299x299 input:

```
max|diff| = 0.006144
corr      = 0.99999913
argmax    TRT=549 timm=549
top5 overlap = 5/5
RESULT: PASS
```

checkpoint is consumed by the builder and nothing is invented.

| Item | Value |
| --- | --- |
| Repository head | `77bbc2c7` |
| Checkpoint | `timm/inception_resnet_v2.tf_in1k` @ `836b07d7d599da247de92f148c7a07fda18afced` |
| Reference | timm 1.0.28 |

### Not Run / Remaining Gaps

- No other Inception-ResNet checkpoint or input resolution was tested.

## Notes For Future Readers

Suggested review order: `plugin.py` first (the residual-block helper and
`_discover_layout` carry all the architecture decisions), then the registration
diffs, which are mechanical and mirror #1155.

If a future Inception-ResNet checkpoint is added and its output is close but
wrong, check the residual scales and `_BN_EPS` first. Both are architecture
constants that no checkpoint records, and both produce a high-correlation
near-miss rather than an obvious failure.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1213, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1169, opened 2026-09-05
feat(timm_hrnet): add timm HRNet image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no HRNet family. HRNet is listed by
wang-xinyu/tensorrtx and is structurally unlike every classifier we already
support: instead of narrowing to a single feature map, it keeps four
resolutions alive in parallel and repeatedly fuses each one into all the
others. No existing family plugin can express that.

## Exit Criteria

- `timm/hrnet_w18.ms_aug_in1k` builds and classifies through the
  `timm_hrnet_image_classification` runtime strategy.

`NotImplementedError`; only the `w18` width is registered, and no segmentation
or pose head is covered.

## Implementation

New `timm_hrnet` family, following the layout the other timm families use: a
directory with one manifest.

The stage grid is read from the checkpoint by `_discover_layout`: module count,
branch count, and per-branch block count for each of stage2/3/4, plus the
`layer1` depth. Transition layers are deliberately *not* counted from their own
keys — an identity transition has no weights, so its index leaves no trace in
the checkpoint. The branch count of the stage a transition feeds is used
instead, and each transition entry is then classified by which keys exist.

The fuse grid follows from the row and column index alone: a higher-index
branch is at lower resolution and is projected by 1x1 convolution then
nearest-upsampled by `2 ** (j - i)`; a lower-index branch is downsampled by a
chain of strided 3x3 convolutions. Only the last convolution in that chain
omits its activation, because its output feeds the sum rather than another
convolution — that asymmetry is the easiest thing to get wrong here and is
commented in `_fuse`.

One other detail worth flagging: the head's downsampling convolutions and its
final projection carry a bias, while every other convolution in the model does
not.

`add_nearest_upsample` is added to this family's graph helpers; the other
convolutional families never need to increase a spatial dimension.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 264.29s
```

random 1x3x224x224 input:

```
max|diff| = 0.010196
corr      = 0.99999844
argmax    TRT=4 timm=4
top5 overlap = 5/5
RESULT: PASS
```

checkpoint is consumed by the builder and nothing is invented. That matters
more than usual here: a wrong fuse-grid index would silently drop or duplicate
a branch, and the name check is what rules that out.

| Item | Value |
| --- | --- |
| Repository head | `77bbc2c7` |
| Checkpoint | `timm/hrnet_w18.ms_aug_in1k` @ `7e2c5583769f54514fd87e3ba9de408e33eaba0f` |
| Reference | timm 1.0.28 |

### Not Run / Remaining Gaps

- Only `hrnet_w18` was tested. The wider variants share the discovery path but
  were not built.

## Notes For Future Readers

Suggested review order: `_discover_layout` and `_fuse` in `plugin.py` carry all
the architecture decisions; the registration diffs are mechanical and mirror
the previous timm families.

If a wider HRNet is added later and fails, the discovery code is the place to
look before the block code: the widths change but the grid logic does not.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1214, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1170, opened 2026-09-05
feat(timm_swin): add timm Swin Transformer image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no Swin Transformer family. Swin is listed by
wang-xinyu/tensorrtx and is the one vision transformer there that `timm_vit`
cannot express: attention is computed inside fixed windows rather than across
all tokens, and alternate blocks shift the window grid by half a window so that
windows overlap between blocks.

## Exit Criteria

- `timm/swin_tiny_patch4_window7_224.ms_in1k` builds and classifies through the
  `timm_swin_image_classification` runtime strategy.

`NotImplementedError`; only the `tiny` variant at 224x224 is registered, and
input sizes that do not divide the window are not handled.

## Implementation

New `timm_swin` family, following the layout the other timm families use: a
directory with one manifest.

The shift is implemented as a cyclic roll of the feature map, which wraps
opposite edges into the same window; a per-window mask then stops those
unrelated positions from attending to each other. Both the relative-position
index and that mask are stored in the checkpoint, so `_attention_bias` folds
them into one additive bias on the host. For a fixed input size that bias is
constant, which keeps the gather and the broadcast out of the engine entirely
and lets the whole attention go through the native TRT attention layer.

The layout is read from the checkpoint: layer count, per-layer depth, embedding
dimension and patch size from the patch-embed convolution, head count from the
relative-position-bias table, and window side from the fourth root of the
relative-position-index element count.

Two graph helpers are new here. `add_roll` builds a cyclic shift from slices and
concatenations, because TensorRT has no roll operator.
`add_global_avg_pool_nhwc` reduces the spatial axes of an NHWC tensor; the
convolutional families all pool NCHW.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 272.11s
```

random 1x3x224x224 input:

```
timm reference: missing=0 unexpected=17
max|diff| = 0.000736
corr      = 0.99999922
argmax    TRT=58 timm=58
top5 overlap = 5/5
RESULT: PASS
```

`unexpected=17` is expected and is a property of the reference, not of this
builder: `relative_position_index` and `attn_mask` are non-persistent buffers in
timm, so `load_state_dict` does not claim them. This plugin does consume them —
they are exactly the tensors `_attention_bias` folds together.

A first version of this change reached corr 0.981 with the argmax still
correct. Reimplementing the plugin's own arithmetic in torch and comparing to
timm block by block showed the logic was exact to float noise, which located
the fault in the TensorRT translation rather than the architecture: `add_roll`
computed its split point with the wrong sign and rolled the opposite way. The
comment on `add_roll` now records that, because the wrong sign still produces a
well-formed tensor of the right shape.

| Item | Value |
| --- | --- |
| Checkpoint | `timm/swin_tiny_patch4_window7_224.ms_in1k` @ `4cc3a7275b50b53a7bec45f32c236ebe64227cff` |
| Reference | timm 1.0.28 |

### Not Run / Remaining Gaps

- Only `swin_tiny` at 224x224 was tested. The builder assumes every stage's
  side length divides the window, which holds at 224 but is not enforced for
  other resolutions; timm pads in that case and this plugin does not.

## Notes For Future Readers

Suggested review order: `_attention_bias` and `_window_attention` in
`plugin.py`, then `add_roll` in `model/model.py`; the registration diffs are
mechanical and mirror the previous timm families.

If a larger Swin or a different input size is added later and the result is
close but wrong, suspect an index or sign error rather than a weight problem.
Everything in this family is reshape and permutation arithmetic, and every
mistake in it produces a correctly shaped tensor.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1216, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1171, opened 2026-09-05
feat(timm_resnest): add timm ResNeSt image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no ResNeSt family. ResNeSt is listed by
wang-xinyu/tensorrtx and cannot reuse `timm_resnet` or `timm_seresnet`: its
bottleneck replaces the 3x3 convolution with split attention, which is a
different gate shape from squeeze-and-excitation, and the `d` variants also
change where the stride lands.

## Exit Criteria

- `timm/resnest50d.in1k` builds and classifies through the
  `timm_resnest_image_classification` runtime strategy.

`NotImplementedError`; only `resnest50d` is registered, and only the `d`
variants' stride placement is implemented.

## Implementation

New `timm_resnest` family, following the layout the other timm families use: a
directory with one manifest.

`_split_attention` is the part worth reading. The convolution emits `radix`
copies of the output channels; the gate is computed from their sum and
normalised with a **softmax across the radix axis**, so the copies compete for
each channel. The SE families this repo already has use a per-channel sigmoid,
where each channel is gated independently — substituting one for the other
gives a plausible wrong answer rather than an error, so the distinction is
commented in place.

Stage depths and the deep stem come from the checkpoint. The stem convolutions
are picked out by tensor rank rather than by index, because the norms
interleaved in the same `nn.Sequential` also have a `.weight`.

Two things the checkpoint cannot record are stated as constants with a comment
saying why, both specific to the `d` variants: a stride never lands on the
split-attention convolution but on an average pool placed after it, and the
shortcut is average-pooled before its 1x1 projection.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 265.92s
```

random 1x3x224x224 input:

```
max|diff| = 0.006126
corr      = 0.99999914
argmax    TRT=21 timm=21
top5 overlap = 5/5
RESULT: PASS
```

checkpoint is consumed by the builder and nothing is invented.

| Item | Value |
| --- | --- |
| Checkpoint | `timm/resnest50d.in1k` @ `2239b0d822126dee3bc7a5357514de35b8f42092` |
| Reference | timm 1.0.28 |

### Not Run / Remaining Gaps

- Only `resnest50d` was tested. A non-`d` ResNeSt would need the stride moved
  back onto the convolution, which this builder does not do.
- Only `cardinality == 1` was exercised. The grouped path in
  `_split_attention` is written generally but no registered checkpoint uses it.

## Notes For Future Readers

Suggested review order: `_split_attention` in `plugin.py`, then `_block`; the
registration diffs are mechanical and mirror the previous timm families.

If a future ResNeSt is added and lands close but wrong, check the two stride
constants first. They are architecture facts that no checkpoint records, and
getting either wrong still produces correctly shaped tensors throughout.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1215, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1173, opened 2026-09-05
feat(timm_convnext): add timm ConvNeXt image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no ConvNeXt family. ConvNeXt is a convolutional
network shaped like a transformer: a large-kernel depthwise convolution stands
in for attention, followed by a LayerNorm and a two-layer MLP with a single
GELU. Neither `timm_resnet` nor `timm_vit` can express that combination.

## Exit Criteria

- `timm/convnext_tiny.in12k_ft_in1k` builds and classifies through the
  `timm_convnext_image_classification` runtime strategy.

`NotImplementedError`; only `convnext_tiny` is registered, and the ConvNeXt-V2
global-response-normalisation block is not implemented.

## Implementation

New `timm_convnext` family, following the layout the other timm families use: a
directory with one manifest.

timm applies the block MLP by transposing to channels-last, running two
`nn.Linear` layers, and transposing back. A Linear over the channel axis is
exactly a 1x1 convolution on NCHW, so `_linear_as_conv` emits it that way and
the whole network stays NCHW — two fewer transposes per block, and the same
arithmetic.

The remaining ConvNeXt-specific pieces are each commented where they appear:
the LayerNorm normalises the **channel** axis rather than the last axis, which
is the transpose of the usual case and still type-checks if you get it wrong;
the downsample puts its norm **before** the strided convolution, unlike every
other downsample in this repo; and the head pools before it normalises, so the
norm sees one value per channel rather than the whole feature map.

Stage count and per-stage depth come from the checkpoint. The layer-scale
`gamma` is applied only when present, so a checkpoint trained without it still
builds.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 265.02s
```

random 1x3x224x224 input:

```
max|diff| = 0.001950
corr      = 0.99999945
argmax    TRT=21 timm=21
top5 overlap = 5/5
RESULT: PASS
```

checkpoint is consumed by the builder and nothing is invented. This is the
check that confirms the MLP rewrite is faithful rather than merely plausible:
`fc1`/`fc2` are consumed as convolution kernels at their original shapes.

| Item | Value |
| --- | --- |
| Checkpoint | `timm/convnext_tiny.in12k_ft_in1k` @ `aa096f03029c7f0ec052013f64c819b34f8ad790` |
| Reference | timm 1.0.28 |

### Not Run / Remaining Gaps

- Only `convnext_tiny` was tested. The wider and deeper variants differ only in
  the discovered depths, but were not built.
- The `use_conv_mlp` variants of ConvNeXt are not distinguished; this builder
  always emits the MLP as convolutions, which is correct for both but was only
  checked against the channels-last checkpoint.

## Notes For Future Readers

Suggested review order: `_block` and the head in `plugin.py`, then
`add_layer_norm_channels` in `model/model.py`; the registration diffs are
mechanical and mirror the previous timm families.

If a future ConvNeXt lands close but wrong, check the LayerNorm axis first.
Normalising the last axis instead of the channel axis produces a correctly
shaped tensor and a plausible output, so nothing structural will catch it.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1211, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1174, opened 2026-09-05
feat(timm_crossvit): add timm CrossViT image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no CrossViT family. CrossViT is listed by
wang-xinyu/tensorrtx and does not fit `timm_vit`: it runs two vision
transformers side by side on the same image at two patch sizes, and they
exchange information through a cross-attention path that has no analogue in a
single-branch ViT.

## Exit Criteria

- `timm/crossvit_9_240.in1k` builds and classifies through the
  `timm_crossvit_image_classification` runtime strategy.

`NotImplementedError`; only `crossvit_9_240` is registered, and the
`crop_scale=True` variants are not handled.

## Implementation

New `timm_crossvit` family, following the layout the other timm families use: a
directory with one manifest.

The two branches see the same image at different resolutions, so the smaller
branch's rescale happens **inside the engine** rather than in preprocessing:
the bundle has one `pixel_values` input, and each branch derives its own size
from its weights. `add_bicubic_resize` sets both the cubic coefficient
(`-0.75`) and `HALF_PIXEL` coordinates explicitly, because TensorRT and PyTorch
choose different defaults and a mismatch shifts every sampled pixel slightly
instead of failing.

The fusion path is the part worth reading. Only class tokens cross between
branches: each branch projects its class token into the other branch's width,
that single token attends over the other branch's patch tokens, and the result
is projected back and re-attached to its own patches. `_cross_attention` has
exactly one query, so its output is one token wide — that asymmetry is what
makes it different from the self-attention in `_block`.

Nearly all geometry is read from the checkpoint, including each branch's input
size: the positional table has one entry per patch plus the class token, and
the patches tile a square, so the size is `sqrt(tokens - 1) * patch`. The head
count is the one thing that cannot be recovered — every projection is square,
so the head split leaves no trace in any tensor shape — and it is stated as a
constant with that reason.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 253.82s
```

random 1x3x240x240 input:

```
max|diff| = 0.002020
corr      = 0.99999935
argmax    TRT=138 timm=138
top5 overlap = 5/5
RESULT: PASS
```

This also covers the in-engine bicubic rescale: the fine branch runs at 240 and
the coarse branch at 224, so a resize that did not match
`torch.nn.functional.interpolate` would show up here.

| Item | Value |
| --- | --- |
| Checkpoint | `timm/crossvit_9_240.in1k` @ `df3051d20d8addd16d03c4230046128c17e32826` |
| Reference | timm 1.0.28 |
| GPU | NVIDIA Blackwell (`2u2g-gen-0295`) |

### Not Run / Remaining Gaps

- Only `crossvit_9_240` was tested. The `crossvit_15_*` and `crossvit_18_*`
  variants share the discovery path but were not built.
- The `_240` checkpoints rescale rather than crop. A `crop_scale=True`
  checkpoint would need a centre crop instead, which this builder does not do.

## Notes For Future Readers

Suggested review order: the fusion loop in `build_engine`, then
`_cross_attention`; the registration diffs are mechanical and mirror the
previous timm families.

The branch indexing in the fusion loop is the easy thing to get wrong: branch
`i`'s projected class token is fused against branch `(i + 1) % n`'s patches and
so runs at *that* branch's width, then is projected back to its own. Every
combination of those indices produces correctly shaped tensors, so only the
numerical comparison catches a mistake.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1217, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

PR #1180, opened 2026-09-06
feat(timm_senet): add timm SENet image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no SENet family. SENet-154 is listed by
wang-xinyu/tensorrtx, and it is the original squeeze-and-excitation network
rather than a ResNet with SE attached, so `timm_seresnet` cannot express it.

## Exit Criteria

- `timm/senet154.gluon_in1k` builds and classifies through the
  `timm_senet_image_classification` runtime strategy.

`NotImplementedError`; only `senet154` is registered.

## Implementation

New `timm_senet` family, following the layout the other timm families use: a
directory with one manifest.

Three things separate this from `timm_seresnet`, and all three are read from
the checkpoint rather than tabulated:

- The stem is three 3x3 convolutions in one `nn.Sequential`, not a single 7x7.
  The convolutions are picked out by tensor rank, because the norms interleaved
  in that same `Sequential` also have a `.weight`, and the last convolution's
  norm is the top-level `bn1` rather than a member of the sequence.
- The bottleneck 3x3 is grouped. The group count is derived from how many
  input channels each filter actually sees rather than hardcoded.
- The projection shortcut is a 3x3 convolution wherever the stage reduces, and
  a 1x1 in the first stage. `_conv` takes the kernel and the padding from the
  weight shape, so both cases fall out without a special case.

The SE gate is a sigmoid over each channel independently. That is worth being
explicit about because ResNeSt's split attention in `timm_resnest` looks
structurally similar but normalises across a radix axis instead; the comment on
`_squeeze_excite` records the distinction.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [x] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Unit and tooling suite:

```
3 failed, 4042 passed, 8 skipped, 23 warnings in 227.16s
```

Lint gates, matching what Source quality runs:

```
ruff check --config ruff.toml <changed .py files>          -> clean
clang-format --dry-run --Werror <changed .cpp/.h files>    -> clean
```

random 1x3x224x224 input:

```
max|diff| = 0.006545
corr      = 0.99999950
argmax    TRT=489 timm=489
top5 overlap = 5/5
RESULT: PASS
```

checkpoint is consumed by the builder and nothing is invented. That is what
confirms the derived group count and the varying shortcut kernel are right: a
wrong group count still produces a correctly shaped tensor.

| Item | Value |
| --- | --- |
| Checkpoint | `timm/senet154.gluon_in1k` @ `86f6d239ed30acbe3aca0cbb5769da2621a4e02c` |
| Reference | timm 1.0.28 |
| GPU | NVIDIA Blackwell (`2u2g-spr-0052`) |

### Not Run / Remaining Gaps

- Only `senet154` was tested. The `legacy_seresnext*` checkpoints share most of
  this structure but were not built.

## Notes For Future Readers

Suggested review order: `_block` in `plugin.py`, then the stem loop in
`build_engine`; the registration diffs are mechanical and mirror the previous
timm families.

If a `legacy_seresnext` checkpoint is added later, the stem is where it will
differ first: those variants use a single 7x7 convolution, so the rank-based
stem discovery will find one convolution instead of three and the top-level
`bn1` will follow it directly.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High


zhenshanx-nv:
Superseded by #1209, which implements this family in the post-cutover layout introduced by #1188. This branch targets `python/tensorrt_model_connect/families/`, which no longer exists, so it cannot be rebased.

Issue #1177, opened 2026-09-05
Cross-family runtime safety and long-context issues found during #1158 review


ruiling-smartbear:
## Context

Reviewing #1158 surfaced several defects in code copied from the existing Llama/model-family templates. I re-audited each call path against current `main` and kept only findings with a concrete failure mechanism. These are not SmolLM3-specific regressions and should not be fixed inside #1158.

#1176 already tracks two other confirmed shared defects: unchecked CUDA device queries in the sparse multinomial policy and GPU fixture tests that can report a false pass.

The counts below are occurrences of the exact code pattern on `main`, not a claim that every family enables the affected optional feature.

## Confirmed findings

- [ ] **Check sampler `cudaMalloc` results transactionally** — allocations are used without checking status. In `ensure_device_buffers`, `capacity_ = keep` runs unconditionally after both `cudaMalloc` calls, so a failed allocation still advances the capacity and is never retried: every later call with `keep <= capacity_` skips reallocation and reuses the bad pointers. The sparse path can then return a stale token when `d_token_id_` is null, and the greedy CUDA kernel can receive a null output pointer. Add allocation-failure injection coverage, throw on allocation failure, and only commit new pointers and capacity after every allocation succeeds. Original finding: https://github.com/NVIDIA/TensorRT-Model-Connect/pull/1158#discussion_r3938042604

- [ ] **Harden temporary FFI kernel loading** — bundle kernel bytes are written with `std::ofstream` to predictable `/tmp/trtmc_kernel_<global_name>.so` paths. Creation is neither exclusive nor no-follow, the complete write is not checked, and files are not removed. A pre-created symlink is followed and concurrent processes using the same global name share the same file before it is loaded. The exact filename pattern appears in 83 family helper files.

[...]

Issue #1176, opened 2026-09-05
Two defects every model family inherited: an unguarded device query and GPU tests that pass without running


ruiling-smartbear:
Both of these came out of the review on #1158. Neither is specific to that family: the code is duplicated per family by design, so a single original decision is now present in 31 to 34 copies. I have not opened a PR for either yet, because the shape is worth agreeing on first.

## 1. `sparse_multinomial_kernel.cu` divides by a discarded device query

```cpp
int device = 0;
cudaGetDevice(&device);                      // return value discarded
cudaDeviceProp props{};                      // zero-initialised
cudaGetDeviceProperties(&props, device);     // return value discarded

const uint32_t blocks_per_sm =
    static_cast<uint32_t>(props.maxThreadsPerMultiProcessor / kDistributionBlockSize);
const uint32_t grid = std::min(
    static_cast<uint32_t>(props.multiProcessorCount) * blocks_per_sm, ...);
const uint64_t total_threads = static_cast<uint64_t>(grid) * kDistributionBlockSize;
const uint64_t counter_offset =
    ((static_cast<uint64_t>(numel) - 1)
     / (total_threads * kGeneratorOffsetsPerCurandCall) + 1)   // divides by total_threads
    * kGeneratorOffsetsPerCurandCall;
```

A failed query leaves `props` as it was, so `blocks_per_sm`, `grid` and `total_threads` are all 0 and the last expression divides by zero. That is undefined behaviour, and on x86 it takes SIGFPE.

Verified rather than inferred:

* On a host with no usable device, `cudaGetDeviceProperties` returns `cudaErrorInsufficientDriver` and writes no byte of the output struct. Probed by passing a buffer filled with `0xAA` and comparing it afterwards; all 4096 bytes came back unchanged. `cudaGetDevice` also fails and leaves `device` at its initialiser, so the code proceeds with device 0.
* Running the same arithmetic over a zeroed struct exits 136.
* 34 families carry this computation. None of them check either return value.

Honest about the reach: getting here needs a CUDA build running where the query fails, and TensorRT would have failed to build or load an engine well before the sampler runs, so the practical exposure is small. It is still undefined behaviour on a path with no guard at all.

**Recommended fix:** check both CUDA calls and fail fast with the returned CUDA error before reading `props` or computing the policy. A device-query failure means this GPU sampler cannot run correctly, so it should surface as a clear runtime error rather than undefined arithmetic.

[...]

PR #1201, opened 2026-09-08
fix(bark): check cudaMalloc status before committing sampler capacity

lukiod:
## Background

`ensure_device_buffers` in `families/bark/runtime/sampler.cpp` didn't check `cudaMalloc`'s return value before committing the new capacity. A failed allocation left the capacity claiming a size the buffer never got, so a later call would reuse a null/stale pointer.

Addresses the cudaMalloc item in #1177.

## Exit Criteria

Every `cudaMalloc` in `ensure_device_buffers` is checked; a failed allocation throws instead of leaving a mismatched capacity, and never leaves an already-succeeded sibling allocation dangling. Success path unchanged. Not covering #1177's other two items or #1176 (both still waiting on maintainer input).

## Implementation

Checks each `cudaError_t`, allocates into local pointers first, and only frees the old buffers and commits the new pointers/capacity once every allocation in the pair succeeds. Same pattern as `core/runtime/tensorrt/trt_module_impl.cpp::ensure_input_buffer`. Only touches `families/bark/runtime/sampler.cpp`.

Second commit: if the indices `cudaMalloc` succeeded but the probabilities one then failed, the old code had already nulled the previous buffers and left the new indices pointer committed to nothing (`entry_capacity_` still 0). Fixed by not touching the existing buffers or capacity until both new allocations succeed.

Third commit addresses @chaofengw-nv's review on #1221. That still left a leak he was right about: `Impl`'s constructor calls `ensure_device_buffers(1, 1)`, and the indices/probabilities pair is committed to members before the token id buffer is allocated - so a failure on that third allocation threw with two allocations already committed, and a constructor that throws never runs its own destructor. Each buffer is now held in a small family-local RAII type, which fixes the class rather than the instance: members that finished constructing are destroyed during unwinding, so whatever was acquired is released no matter which allocation fails. The explicit destructor is now redundant and removed.

### Change categories

### Commands and Results

CUDA and TensorRT are installed now. Built and ran the actual repository target:

```
$ cmake --build build --target test_bark_pipeline
$ ctest -R bark_pipeline --output-on-failure
1/1 Test #1: bark_pipeline ....................   Passed    0.34 sec
100% tests passed out of 1
```

`test_bark_pipeline` exercises the real `sample_rows` path end to end (semantic and coarse token generation), which calls `ensure_device_buffers` on the actual GPU.

New in-repo test, `test_bark_sampler_alloc`, links the real sampler translation unit against CPU CUDA stubs and fails each of the three constructor allocations in turn. It checks that a failed construction releases everything it acquired, that a later construction still succeeds, and that a failed resize leaves the existing buffers intact and the sampler usable:
```
$ ctest
1/2 Test #1: bark_pipeline ....................   Passed
2/2 Test #2: bark_sampler_alloc ...............   Passed
100% tests passed out of 2
```
It needs no GPU and links no cudart, so it runs in the CPU tier. Verified it catches the bug: against `72fba75` it fails with "failed construction must release every allocation it made" on the third allocation only - matching exactly what @chaofengw-nv reported from his own stub run.

Also kept the standalone GPU harnesses from before as targeted evidence for the allocation-failure paths, since `test_bark_pipeline` only exercises the success path:

Original single-allocation-failure case:
```
$ ./test_ensure_buffers
caught expected: Unable to allocate bark sampler index buffer
ALL CHECKS PASSED on real GPU
```

Second-allocation-failure / leak case, forced by starving the device down to room for exactly one of the two equal-sized buffers:
```
[buggy] threw=1 entry_capacity_=0 device_indices_=0x10007600000 device_probabilities_=(nil)
[buggy] free before=99483648 after=32374784 (memory still held despite entry_capacity_==0: YES - LEAK)
[fixed] threw=1 entry_capacity_=0 device_indices_=(nil) device_probabilities_=(nil)
[fixed] free before=99483648 after=99483648 (no memory held: confirmed, no leak)
ALL CHECKS PASSED: buggy path reproduces the leak, fixed path does not.
```

Also re-checked the ordinary path on the fixed version (grow 0->100, no-op shrink to 50, grow to 500, real `cudaMemcpy` round trip): unchanged.

`clang-format --dry-run --Werror` on the patched file: clean.

Repo head rebased onto `54d6d566f` (upstream/main). GPU: GTX 1650, driver 610.57.04, CUDA UMD 13.3. g++ 16.2.1. `cuda` 13.3.1-1 and `nlohmann-json` 3.12.0-2 via pacman; TensorRT 11.2.1 headers (from the matching `v11.2` tag of NVIDIA/TensorRT, Apache-2.0) plus the real `libnvinfer`/`libnvinfer_plugin`/`libnvonnxparser` binaries from the `tensorrt-cu13-libs` wheel.

### Not Run / Remaining Gaps

`test_bark_pipeline` runs the fine/codec stages without real TRT engines (falls back to zero codebooks / a simple waveform, printed plainly by the pipeline itself), so it does not exercise a TRT-engine-backed run end to end - that needs real model weights, out of scope here.

Re-grepped `families/` for the same unchecked-`cudaMalloc` shape against the current head: this repo's `bart`, `marian`, `m2m_100`, `t5`, `whisper`, and `flux` all still have genuinely unchecked `cudaMalloc` calls (encoder-decoder cross-attention KV cache allocation in the first five, a shared matmul workspace in `flux`), same class of bug as this PR fixed in `bark`. That is a materially larger footprint than an earlier version of this section claimed - filing a separate issue for it rather than folding a six-family fix into this one.

## Contributor Self-Review

Change is family-local to `families/bark/runtime/sampler.cpp` only, no shared-core or cross-family touch. Both commits are DCO signed off. The real `test_bark_pipeline` CTest target now passes on this head.

## Notes For Future Readers

If this pattern shows up in more families once other pending work lands, this is a direct template to repeat.

### Risk level
- [x] Low

Single-file, family-local error-handling fix on an existing allocation path; no shared-core, public API, or ABI change; no new behavior on the success path.

### DIFF
--- a/families/bark/runtime/CMakeLists.txt
+++ b/families/bark/runtime/CMakeLists.txt
@@ -65,4 +65,22 @@ if(TRTMC_BUILD_TESTS)
   )
   add_test(NAME bark_pipeline COMMAND test_bark_pipeline)
   set_tests_properties(bark_pipeline PROPERTIES SKIP_RETURN_CODE 77 LABELS gpu)
+
+  # Compiles the sampler against CPU CUDA stubs defined in the test, so
+  # allocation failures can be injected without a GPU or cudart.
+  add_executable(test_bark_sampler_alloc
+    ${PROJECT_SOURCE_DIR}/families/bark/tests/cpp/test_bark_sampler_alloc.cpp
+    sampler.cpp
+  )
+  target_include_directories(test_bark_sampler_alloc PRIVATE
+    ${PROJECT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}/core/runtime/include
+  )
+  target_include_directories(test_bark_sampler_alloc SYSTEM PRIVATE
+    ${TRTMC_CUDA_INCLUDE_DIR}
+  )
+  target_compile_options(test_bark_sampler_alloc PRIVATE
+    -Wall -Wextra -Wpedantic
+  )
+  add_test(NAME bark_sampler_alloc COMMAND test_bark_sampler_alloc)
 endif()
--- a/families/bark/runtime/sampler.cpp
+++ b/families/bark/runtime/sampler.cpp
@@ -69,18 +69,31 @@ FilteredDistribution build_filtered_distributions(const float* logits, int32_t r
     return distribution;
 }
 
+template <typename T>
+class DeviceBuffer {
+  public:
+    DeviceBuffer() = default;
+    ~DeviceBuffer() { cudaFree(ptr_); }
+    DeviceBuffer(const DeviceBuffer&) = delete;
+    DeviceBuffer& operator=(const DeviceBuffer&) = delete;
+
+    T* get() const { return ptr_; }
+
+    void reset(T* pointer) {
+        cudaFree(ptr_);
+        ptr_ = pointer;
+    }
+
+  private:
+    T* ptr_{nullptr};
+};
+
 } // namespace
 
 class BarkSampler::Impl {
   public:
     explicit Impl(void* stream) : stream_(static_cast<cudaStream_t>(stream)) {
         ensure_device_buffers(1, 1);
-    }
-
-    ~Impl() {
-        cudaFree(device_indices_);
-        cudaFree(device_probabilities_);
-        cudaFree(device_token_ids_);
     }
 
     void reset(int64_t seed) {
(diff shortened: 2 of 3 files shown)


lukiod:
Correction on my last update: I said only bark had this unchecked-cudaMalloc shape, but I hadn't actually re-run that grep against the rebased head. Just did - bart, marian, m2m_100, t5, whisper, and flux all still have it too. Opening a separate issue for those rather than folding a six-family fix in here.

PR #1214, opened 2026-09-08
feat(timm_hrnet): add timm HRNet image-classification family

zhenshanx-nv:
## Background

TensorRT-Model-Connect has no HRNet family. HRNet is listed by
wang-xinyu/tensorrtx and is structurally unlike every classifier here: instead
of narrowing to a single feature map it keeps four resolutions alive in parallel
and repeatedly fuses each into all the others.

This replaces #1169, which was written against the family layout removed by the
cutover in #1188. The architecture work and the parity result carry over; the
surrounding code is rewritten for the current layout.

## Exit Criteria

- `timm/hrnet_w18.ms_aug_in1k` builds and classifies through the family `build`
  entrypoint.
- `python tools/test_impact.py --validate` stays valid with no edit outside
  `families/timm_hrnet/`.

Non-goals: quantization, tensor and context parallelism, and batch sizes above
one all raise; only `hrnet_w18` is registered, and no segmentation or pose head
is covered.

## Implementation

One directory, `families/timm_hrnet/`, following
`website/docs/extend/add-model-family.md`.

The stage grid is read from the checkpoint: module count, branch count, and
per-branch block count for each stage, plus the `layer1` depth. Transitions are
deliberately **not** counted from their own keys, because an identity transition
carries no weights and leaves no trace; the branch count of the stage a
transition feeds is authoritative instead, and each entry is then classified by
which keys exist.

The fuse grid follows from the row and column index alone: a higher-index branch
is at lower resolution and is projected by 1x1 then nearest-upsampled by
`2 ** (column - row)`; a lower-index branch is downsampled by a chain of strided
3x3 convolutions. **Only the last convolution in that chain omits its
activation**, because its output feeds the sum rather than another convolution.
That asymmetry is the easiest thing to get wrong and is commented in `_fuse`.

One folding detail specific to this family: HRNet's head convolutions carry a
bias of their own on top of the batch norm, so `_fold` absorbs both. Dropping
the convolution bias shifts each channel by a constant and still ranks
plausibly, so a unit test pins the algebra.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

Family unit tests:

```
python -m pytest families/timm_hrnet/tests/test_model.py -q
6 passed in 0.46s
```

Lint gates:

```
ruff check --config ruff.toml families/timm_hrnet/     -> All checks passed
clang-format --dry-run --Werror <runtime and test C++> -> clean
```

Numerical parity, fp32 build against timm on identical weights and one fixed
random 1x3x224x224 input:

```
max|diff| = 0.011644
corr      = 0.99999888
argmax    TRT=4 timm=4
top5 overlap = 5/5
RESULT: PASS
```

`missing=0 unexpected=0` matters more than usual here: a wrong fuse-grid index
would silently drop or duplicate a branch, and the name check is what rules that
out.

| Item | Value |
| --- | --- |
| Repository head | `b5ae14b0` |
| Checkpoint | `timm/hrnet_w18.ms_aug_in1k` @ `7e2c5583769f54514fd87e3ba9de408e33eaba0f` |
| Reference | timm 1.0.28 |
| GPU | NVIDIA Blackwell (`2u2g-spr-0084`) |

### Not Run / Remaining Gaps

- The E2E case was not run locally; it needs `TRTMC_BINARY` and the CI image
  cache. It is covered by the gate.
- Only fp32 was compared against the reference.
- Only `hrnet_w18` was built. The wider variants share the discovery path but
  were not run.

## Contributor Self-Review

Reviewed the full diff before marking ready. The parity number, the unit
tests, and the lint gates in the Validation section were all run on this
exact revision.

## Notes For Future Readers

Suggested review order: `_layout` and `_fuse` in `model.py`.

If a wider HRNet is added later and fails, look at the discovery code before the
block code: the widths change but the grid logic does not.

### Risk level

- [x] Low
- [ ] Medium
- [ ] High

The change is additive and confined to one new family directory.


zhenshanx-nv:
Flagging a mistake in this merged PR, for the record.

Besides the `families/timm_hrnet/` directory and its benchmark release entry,
this PR also carried four files that its title and description do not mention:

- `core/runtime/include/trtmc/task.h` - adds `DetectionBox`,
  `ObjectDetectionResult` and the `IObjectDetection` interface
- `apps/cli/cli.h`, `apps/cli/cli.cpp` - add a `detect` command
- `apps/benchmark/trtmc_benchmark/task_adapters.py` - adds the
  `object_detection` task default

Those belong to #1220, which I had open for exactly that change. They reached
this branch by accident: I was editing them in the same working tree that my
merge automation drives, and when the automation reset that tree onto the
target branch git carried the uncommitted edits across, where an `add -A`
swept them into this commit.

The code is not in question - it is the same change that was reviewed on #1220,
builds clean, and nothing yet implements the new interface, so behaviour is
unchanged. What is wrong is the provenance: approving this PR meant approving a
new public runtime interface and a new CLI command that were not described here.

I am closing #1220 as already landed rather than reverting, since a revert would
churn `main` to remove an interface no family uses yet. Happy to do the revert
and re-land properly instead if a maintainer prefers that - say the word.

My automation now runs in a worktree separate from where I edit, so this
particular failure cannot recur.

Issue #1221, opened 2026-09-09
Six families discard cudaMalloc's return value like bark did before #1201


lukiod:
`bark/runtime/sampler.cpp`'s `cudaMalloc` calls used to discard the return value entirely (fixed in #1201). The same shape exists in six other families, verified against current `main` rather than assumed:

- `bart/runtime/plugin.cpp:163-164` (`BartPipeline` constructor, `cross_k_ptrs_`/`cross_v_ptrs_`)
- `marian/runtime/plugin.cpp:193-194` (constructor, same pair) and `:282` (`enc_mask_device_`, lazy allocation)
- `m2m_100/runtime/plugin.cpp:172-173` (constructor, same pair)
- `t5/runtime/plugin.cpp:192-193` (constructor, same pair) and `:297` (`enc_mask_device_`, lazy allocation)
- `whisper/runtime/pipeline.cpp:58-59` (constructor, same pair)
- `flux/runtime/gpu_matmul.cpp:32` (`ensure_buf`, shared matmul scratch workspace)

None of these check the `cudaError_t`. In the five encoder-decoder families this runs once per decoder layer in the constructor - on an out-of-memory allocation failure the pointer stays null, the constructor returns as if construction succeeded, and the first `cudaMemcpyAsync`/kernel bind into that buffer either errors opaquely or silently no-ops depending on the driver, well after the point where the real cause (allocation failure) was known and could have been reported clearly. `flux`'s `ensure_buf` is a shared scratch buffer reused across matmul calls, same failure mode.

Each site's destructor already guards on a non-null pointer before `cudaFree`, so committing `nullptr` on failure and throwing (matching `bark`'s fix in #1201: `core/runtime/tensorrt/trt_module_impl.cpp::ensure_input_buffer`'s existing convention - check `cudaError_t`, throw `std::runtime_error` on failure) is a direct, minimal fix per site, no shared abstraction needed.

Not fixed here since each is a separate family-owned file and this repo's convention is one family per PR. Filing so the pattern is tracked in one place rather than found six more times independently.

chaofengw-nv:
Thanks @lukiod for identifying this pattern and submitting the family-local fixes. I checked the reported allocation sites and reviewed #1222 (`55e784e27`), #1223 (`3cc896ab8`), and the related Bark fix #1201 (`72fba7581`). The unchecked allocations are real, and checking their return values is the right direction.

There is one remaining exception-safety issue to address before merging:

[...]

PR #1232, opened 2026-09-10
fix(m2m_100): check cudaMalloc status for cross-attention buffers

lukiod:
## Background

`M2M100Pipeline`'s constructor allocates one cross-attention key and value buffer per decoder layer but never checked `cudaMalloc`'s result. Same shape as bark (#1201), bart (#1222) and whisper (#1223), all still open, tracked together in #1221.

## Exit Criteria

Every `cudaMalloc` in the constructor is checked, and a failure releases every allocation the constructor already made rather than leaking it. Success path unchanged.

## Implementation

Two commits.

The first checks each `cudaError_t` and throws on failure, matching `core/runtime/tensorrt/trt_module_impl.cpp`'s existing convention.

The second addresses @chaofengw-nv's review on #1221: checking the status was only half the fix. The buffers lived in `std::vector<void*>`, and destroying that vector does not free the device allocations its elements point at - and a constructor that throws never runs its own destructor - so a failure on any layer after the first leaked everything allocated before it. Each buffer is now held in a family-local `DeviceBuffer` (`families/m2m_100/runtime/device_buffer.h`), so the vector members release what they hold during unwinding. The explicit destructor loop is now redundant and removed.

### Change categories

### Commands and Results

Built the real targets and ran the family's tests:
```
$ cmake --build build --target trtmc_model_m2m_100 test_m2m_100_runtime_config test_m2m_100_tokenizer_schema test_m2m_100_cross_kv_alloc
$ ctest
3/3 Test #3: m2m_100_cross_kv_alloc ...........   Passed
100% tests passed out of 3
```

`test_m2m_100_cross_kv_alloc` is new. `M2M100Pipeline` needs live TensorRT modules to construct, so the test drives the same allocation loop over the same owning type against CPU CUDA stubs, failing each of the eight allocations in turn and asserting which buffer the reported error names. It needs no GPU and links no cudart, so it runs in the CPU tier.

Verified it actually catches the bug rather than just passing: against the previous raw-pointer shape it reports 8 failed checks, matching the "leaks from the second allocation onward" finding on #1221.

`clang-format --dry-run --Werror` on the changed files: clean.

Rebased onto `1f7fc4156` (upstream/main). GPU: GTX 1650, driver 610.57.04. g++ 16.2.1. CUDA 13.3.1, TensorRT 11.2.1.

### Not Run / Remaining Gaps

No real M2M-100/NLLB checkpoint or engine on hand, so the full pipeline was not exercised end to end - only the buffer ownership this PR changes.

## Contributor Self-Review

Single family, no shared-core or cross-family touch. Both commits DCO signed off.

## Notes For Future Readers

`t5`, `marian` and `flux` still carry the same pattern, tracked in #1221.

### Risk level
- [x] Low

Error-handling and ownership change on an existing allocation path; no public API, ABI, or success-path behavior change.

### DIFF
--- a/families/m2m_100/runtime/CMakeLists.txt
+++ b/families/m2m_100/runtime/CMakeLists.txt
@@ -57,4 +57,21 @@ if(TRTMC_BUILD_TESTS)
   target_link_libraries(test_m2m_100_tokenizer_schema PRIVATE trtmc_model_m2m_100)
   target_compile_options(test_m2m_100_tokenizer_schema PRIVATE -Wall -Wextra -Wpedantic)
   add_test(NAME m2m_100_tokenizer_schema COMMAND test_m2m_100_tokenizer_schema)
+
+  # CPU CUDA stubs live in the test, so allocation failures can be injected
+  # without a GPU or cudart.
+  add_executable(test_m2m_100_cross_kv_alloc
+    ${PROJECT_SOURCE_DIR}/families/m2m_100/tests/cpp/test_m2m_100_cross_kv_alloc.cpp
+  )
+  target_include_directories(test_m2m_100_cross_kv_alloc PRIVATE
+    ${PROJECT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}/core/runtime/include
+  )
+  target_include_directories(test_m2m_100_cross_kv_alloc SYSTEM PRIVATE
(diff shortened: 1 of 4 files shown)


chaofengw-nv:
Sorry for the repeated close/reopen cycles on this PR. Community CI should be working correctly now; the PR has been merged. Thanks for your contribution！

PR #1233, opened 2026-09-10
fix(t5): check cudaMalloc status for cross-attention and encoder mask buffers

lukiod:
## Background

`T5Pipeline` discarded the result of every `cudaMalloc`: the constructor allocates a cross-attention key and value buffer per decoder layer, and `generate()` lazily allocates the encoder mask. A failed allocation left a null pointer that the pipeline then bound to the decoder as if it were real memory. Tracked in #1221; same shape already merged for bark (#1201), bart (#1222) and whisper (#1223).

## Exit Criteria

Every `cudaMalloc` is checked, and a failure releases whatever the call already acquired rather than leaking it. Success path unchanged.

## Implementation

Checking the status alone would not have been enough. The buffers lived in `std::vector<void*>`, destroying that vector does not free the device allocations its elements point at, and a constructor that throws never runs its own destructor - so any layer failing after the first leaked everything allocated before it.

Each buffer, including the lazily-allocated encoder mask, is now held in a family-local `DeviceBuffer` (`families/t5/runtime/device_buffer.h`), so the members release what they hold during unwinding. The explicit destructor is now redundant and removed. Only touches `families/t5/`.

### Change categories

### Commands and Results

```
$ cmake --build build --target trtmc_model_t5 test_t5_runtime_config test_t5_cross_kv_alloc
$ ctest
1/2 Test #1: t5_runtime_config ................   Passed
2/2 Test #2: t5_cross_kv_alloc ................   Passed
100% tests passed out of 2
```

`test_t5_cross_kv_alloc` is new. `T5Pipeline` needs live TensorRT modules to construct, so the test drives the same allocation loop over the same owning type against CPU CUDA stubs, failing each of the eight allocations in turn and asserting which buffer the reported error names. It needs no GPU and links no cudart, so it runs in the CPU tier.

Verified it catches the bug rather than just passing: against the previous raw-pointer shape it reports 8 failed checks.

`clang-format --dry-run --Werror` on the changed files: clean.

Branched from `911d56607` (upstream/main). GPU: GTX 1650, driver 610.57.04. g++ 16.2.1. CUDA 13.3.1, TensorRT 11.2.1.

### Not Run / Remaining Gaps

No real T5 checkpoint or engine on hand, so the full pipeline was not exercised end to end - only the buffer ownership this PR changes.

## Contributor Self-Review

Single family, no shared-core or cross-family touch. DCO signed off.

## Notes For Future Readers

`marian` and `flux` are the last two from #1221; branches are ready and will follow.

### Risk level
- [x] Low

Error-handling and ownership change on an existing allocation path; no public API, ABI, or success-path behavior change.

### DIFF
--- a/families/t5/runtime/CMakeLists.txt
+++ b/families/t5/runtime/CMakeLists.txt
@@ -47,4 +47,21 @@ if(TRTMC_BUILD_TESTS)
   target_link_libraries(test_t5_runtime_config PRIVATE trtmc_model_t5)
   target_compile_options(test_t5_runtime_config PRIVATE -Wall -Wextra -Wpedantic)
   add_test(NAME t5_runtime_config COMMAND test_t5_runtime_config)
+
+  # CPU CUDA stubs live in the test, so allocation failures can be injected
+  # without a GPU or cudart.
+  add_executable(test_t5_cross_kv_alloc
+    ${PROJECT_SOURCE_DIR}/families/t5/tests/cpp/test_t5_cross_kv_alloc.cpp
+  )
+  target_include_directories(test_t5_cross_kv_alloc PRIVATE
+    ${PROJECT_SOURCE_DIR}
+    ${PROJECT_SOURCE_DIR}/core/runtime/include
+  )
+  target_include_directories(test_t5_cross_kv_alloc SYSTEM PRIVATE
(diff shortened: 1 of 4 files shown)


yifeif-nv:
Hi @lukiod thanks for the contribution. The internal CI passed and this looks good. Merging it now

PR #1181, opened 2026-09-06
feat(detr): add DETR object-detection family

husizhtu:
## Background

Add native support for the DETR object-detection family, targeting `facebook/detr-resnet-50`, under the family-isolated architecture. Current main already provides the shared `IObjectDetection` contract and `trtmc detect` CLI command, so this PR implements that existing contract for DETR rather than adding new shared API.

## Exit Criteria

- New family is discoverable as `detr`.
- Python builder resolves the checkpoint, loads weights, and builds a TensorRT engine via the Network API.
- C++ runtime DSO exposes `trtmc_create_family` and returns `ObjectDetectionResult`.
- E2E manifest with a pinned HF revision is present.
- `tools/model_ci.py validate` and `tools/test_impact.py --validate` pass.
- Non-goals: quantization, tensor-parallel, and auxiliary-loss training paths.

## Implementation

- Added the DETR family under `families/detr/`:
  - Python TensorRT Network API builder
  - checkpoint loader for `facebook/detr-resnet-50`
  - C++ preprocessing and post-processing runtime implementing the existing `IObjectDetection` contract
  - family-owned E2E tests and manifests
- Added focused shared-contract coverage for the existing `IObjectDetection` API and `trtmc detect` JSON output.
- Updated the public Task API documentation for object detection.
- Added the shared test fixture to `ASSET_LICENSES.md`.
- Rebased onto current `main`; no central registry or sibling-family dependency is added.`n- Added `timm==1.0.28` to the family requirements for the Hugging Face DETR reference.

### Change categories

- [x] Model or runtime behavior
- [ ] Public API
- [ ] ABI
- [x] Bundle or artifact format
- [x] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

- `python tools/test_impact.py --validate` -> valid.
- `python -m tools.model_ci validate` -> valid.
- Community CPU checks: source quality, docs, unit, ownership/impact, and required all pass on the rebased head.
- C++ sources pass `clang-format --dry-run --Werror` with clang-format 22.1.8.

### Hardware, Environment, and Revisions

- Repository head: current PR head commit.
- Checkpoint: `facebook/detr-resnet-50` at revision `1d5f47bd3bdd2c4bbfa585418ffe6da5028b4c0b`.
- Local validation environment: Windows with community CPU tooling.
- Not run: TensorRT engine build and GPU E2E inference.

### Not Run / Remaining Gaps

[...]

PR #1242, opened 2026-09-10
feat(detr): add the DETR object-detection family

zhenshanx-nv:
## Background

DETR is the transformer detector from wang-xinyu/tensorrtx that Model Connect
did not cover. It is the first non-YOLO detector in the repository and the
second user of the `object_detection` task, which #1228 introduced.

## Exit Criteria

- `facebook/detr-resnet-50` builds and agrees with transformers on the
  detections it reports.
- The native runtime path returns the same detections as the builder does.
- The engine's fixed input size is explicit rather than implied.

Non-goals: quantization, batching above one, tensor parallelism, dilated
backbones, learned position embeddings, and the panoptic segmentation head.

## Implementation

Adds `families/detr/`. DETR predicts a fixed set of boxes directly: a ResNet
backbone produces one feature map, a transformer encoder attends over its
pixels, and a decoder turns a fixed set of learned queries into one box and one
label each. There is no anchor grid and no non-maximum suppression, so the set
is the output and the engine only has to rank it.

Two decisions are worth calling out:

- The engine is built for **one fixed input size**, so the sine position
  embedding is a constant and is computed on the host instead of in the graph.
  `--image-height` / `--image-width` name that size; the default is 800x800.
- Runtime preprocessing is a **plain resize**, not a letterbox. DETR normalises
  its boxes to its own input, so a plain resize keeps the mapping back to source
  pixels one multiply per axis. Padding would instead require the position
  embedding to know which pixels are real.

Position is added to the attention query and key but never to the value, in
both the encoder and the decoder cross-attention. The backbone norms are frozen
and fold into the convolution before them using DETR's own hard-coded 1e-5.

The layout comes from the checkpoint: backbone stage depths from the
`layer<stage>.<block>` keys, encoder and decoder depths from their
`layers.<index>` keys, and the class count from the classifier weight.

### Change categories

### Commands and Results

**Parity**, fp32 and fp16 engines against transformers on the same tensor, real
photograph, comparing every detection at or above the 0.5 operating threshold:

| Input | Precision | Detections | Labels | Max score diff | Max box diff |
| --- | --- | --- | --- | --- | --- |
| 800x800 | fp32 | 2 = 2 | match | 0.0017 | 6.4e-5 |
| 640x640 | fp32 | 2 = 2 | match | 0.0045 | 6.6e-5 |
| 1024x1024 | fp32 | 2 = 2 | match | 0.0050 | 7.3e-5 |
| 800x800 | fp16 | 2 = 2 | match | 0.0036 | 7.0e-4 |

Box differences are normalised to the input, so 6.4e-5 is about 0.05 px at 800.
Tolerances are precision-dependent: what decides a detection is the label, the
count above the threshold, and the box, and a half-percent score move on a 0.88
query cannot flip a 0.5 cut.

**Native runtime end to end**, `pytest families/detr/tests/test_e2e.py
--e2e-model detr` with the built CLI and runtime -> **1 passed**. The engine's
detections match the reference to under one source pixel per box edge.

**C++ seam**, `test_detr_image_preprocess` -> 5 tests pass, covering the
interleaved-to-planar layout, per-channel normalisation, the resize, and both
rejection paths.

**Unit**: `pytest families/detr/tests/test_model.py -q` -> 16 passed.

**Repository**: `pytest families/detr/tests/test_model.py tools/tests
apps/benchmark/trtmc_benchmark/tests -q -m "not gpu and not trt"` -> 320 passed.

Source quality: `ruff check`, `ruff format --check`, `clang-format` all clean.

- Repository head: branched from `upstream/main` at `5c0bedd8`.
- GPU: NVIDIA A100 80GB PCIe, CUDA 13.0, TensorRT 11.1.0.106.
- transformers 5.2.0, torch 2.12.0 (CPU build; the reference runs on CPU and
  the engine on the GPU, so this is fp32 CPU against fp32 GPU).
- Checkpoint: `facebook/detr-resnet-50` at
  `1d5f47bd3bdd2c4bbfa585418ffe6da5028b4c0b`.

### Not Run / Remaining Gaps

- Only `facebook/detr-resnet-50` was exercised. `detr-resnet-101` should work
  from the same code path, since the depth is read from the checkpoint, but it
  was not built.
- Accuracy against COCO was not measured; the evidence here is agreement with
  the reference implementation, not a mAP number.
- The panoptic checkpoints are out of scope and are rejected by the identity
  check rather than mis-built.

## Contributor Self-Review

## Notes For Future Readers

Three defects in this change were found only by building the native runtime and
running it, and none were in the model. The builder-side parity above was clean
at correlation 1.0 the entire time, because it never touched the C++ path:

1. The preprocessing seam read the CLI's **interleaved** RGB as **planar**. The
   engine received a scrambled image. The same bug shipped in YOLOv10 and was
   caught there by CI (#1228); its seam test could not see it because it used a
   constant-valued image, where the two layouts are identical. Both seam tests
   now use distinct per-channel values.
2. The C++ seam test did not compile: it still carried YOLOv10's letterbox
   assertions from the scaffold, and nothing had built it with tests enabled.
3. The e2e indexed `boxes` as nested. The CLI writes one flat array, four
   values per detection.

The e2e resizes its reference with the same bilinear the runtime seam uses,
rather than PIL's. Measured across the whole frame the two differ by a uniform
~1.1e-3 per pixel with no spatial structure, and the detector amplifies that
into roughly two pixels of box. Feeding both sides the same pixels keeps the
one-pixel box tolerance meaningful; widening the tolerance instead would have
hidden a real difference behind a looser number. The seam's own resize is
covered by its C++ test.

Third-party provenance: DETR is a transformers architecture; the checkpoint is
the published `facebook/detr-resnet-50` repository.

### Risk level

- [x] Low

New family behind its own identity check. It adds one entry to the shared
release suite and changes no existing family.


zhenshanx-nv:
Closing this as a duplicate of #1181, which was opened on 6 September, four
days before this one. I did not check for an existing DETR pull request before
starting, which is my mistake.

#1181 is also the better implementation, so there is nothing here worth
salvaging into it:

- Its preprocessing preserves aspect ratio with a shortest-edge / longest-edge
  resize, matching `DetrImageProcessor`. This one resized straight to a square
  and distorted the image.
- Its end-to-end test drives the real `DetrImageProcessor` and
  `post_process_object_detection` rather than a hand-rolled reference.
- It avoids all three defects I had to fix here: the preprocessing seam reading
  the CLI's interleaved RGB as planar, the seam test not compiling, and the
  end-to-end test indexing the CLI's flat `boxes` array as if it were nested.

The one thing this branch has that #1181 does not is the post-cutover family
layout (`checkpoint.py` and a flat `graph.py`, no `config.py` / `weights/`),
since #1181 predates that change. That is a rebase on #1181, not a reason to
prefer this branch.

@husizhtu, apologies for the duplicated effort. If the failing internal gate is
useful to chase, I have a built DETR runtime and a GPU node here and am happy
to reproduce it against your branch; say the word.

PR #1269, opened 2026-09-11
fix(llama): preserve full eos_token_id list through to the runtime

roma5087:
## Background

HF `config.json`/`generation_config.json` may declare `eos_token_id` as either
a single id or a list of stop-token ids (all Llama 3.1+ checkpoints, and
`openbmb/MiniCPM5-2B`, whose config has `eos_token_id: [1, 130073]`). The
`families/llama` build path only ever kept a single scalar
`ModelConfig.eos_token_id`, so a multi-id checkpoint was silently truncated to
its first id when writing `runtime.json`, and the C++ runtime loader
(`families/llama/runtime/plugin.cpp`) would reject a list value outright,
crashing bundle load with `llama runtime.json has invalid 'eos_token_id'`.
Truncating to the first id (an earlier version of this fix did that) is also
semantically wrong for the flagship case: Llama 3.1-Instruct's real per-turn
stop token (`<|eot_id|>`) is the *last* id in its list, not the first, so that
approach would silently break generation stopping instead of crashing.

No originating issue is linked — per `CONTRIBUTING.md`, an issue is expected
before investing in a large/cross-cutting/design-unclear change; this is
scoped to one family with a clear, already-validated approach, so no issue was
opened first.

## Exit Criteria

- A checkpoint with a list-valued `eos_token_id` builds and its bundle loads
  and runs without crashing.
- All ids in the list are honored as valid stop tokens at generation time, not
  just the first.
- Ordinary single-scalar `eos_token_id` checkpoints continue to build and run
  unchanged (aside from the wire-format wrapping described below).
- Non-goals: this PR does not fix the parallel `bos_token_id`/`pad_token_id`
  missing-vs-zero bug (see Notes), and does not fix an unrelated,
  pre-existing TinyLlama generation bug discovered during regression testing
  (see Notes) — both are intentionally out of scope.

## Implementation

Preserve the full `eos_token_id` list end-to-end instead of collapsing it to
one id:

- `families/llama/config.py`: add `ModelConfig.eos_token_ids: tuple[int, ...]`
  and a shared `_as_id_list()` helper that normalizes a missing value, a
  scalar, or a list into a tuple, checking for `None` explicitly (not
  truthiness), so a legitimate id of `0` survives.
- `families/llama/model.py`: `runtime.json`'s `"eos_token_id"` is now always
  emitted as a JSON array, for both the `config.json`-derived path and the
  `generation_config.json` override path (the override path previously had no
  empty-list fallback at all; it now falls back to `[-1]`).
- `families/llama/runtime/plugin.cpp`: `RuntimeConfig::eos_token_id` (scalar)
  becomes `eos_token_ids` (`vector<int32_t>`), parsed as a JSON array with an
  explicit empty-vector guard, and wired into
  `LlamaTextGenConfig::id_eos_ids` — the pipeline's existing multi-stop-token
  mechanism (`pipeline.cpp`'s `normalize_eos_token_ids`, `sampler.cpp`'s
  `is_stop_token`/`build_sampling_params`), which already existed and was
  already unit-tested, but was never being fed more than one id.

Affected: `families/llama` only (build path + native runtime plugin). No
other family is touched.

**Bundle/artifact format change**: `runtime.json`'s `eos_token_id` field
changes from sometimes-scalar to always-array. This is a breaking change to
the bundle wire format (see Notes for details) — no public API or ABI change.

### Change categories

- [ ] Public API
- [ ] ABI
- [x] Bundle or artifact format
- [ ] Dependencies
- [ ] Documentation only
- [ ] CI or developer tooling

### Commands and Results

```
pytest families/llama/tests/test_config.py            # 4 passed
pytest families/llama/tests/test_runtime_config.py     # 3 passed (new)
pytest families/llama/tests/ --ignore=families/llama/tests/test_e2e.py
                                                        # 16 passed (incl. real-TensorRT
                                                        #  test_builder_variants.py, test_rope_scaling.py)
pytest families/llama/tests/test_e2e.py --e2e-model=minicpm5-2b -s
                                                        # 1 passed, 7 skipped
                                                        # (native output compared against a real
                                                        #  transformers.generate() reference via the
                                                        #  family's normalized-edit-distance oracle)
```

Regression check (not an automated test, manual verification): rebuilt
`tinyllama-1.1b` (plain scalar `eos_token_id: 2`) with this fix, confirmed
`runtime.json` correctly contains `"eos_token_id":[2]` and the native runtime
loads and runs it without error (see Notes for a separate, pre-existing,
unrelated generation-quality issue found while doing this check).

- Tested head: this branch, rebased onto `upstream/main` (fast-forward, zero
  divergent commits at time of testing)
- GPU: NVIDIA A40, 48GB VRAM, driver 580.159.03 (CUDA forward-compatibility
  mode, CUDA 13.3 userspace)
- Container: `nvcr.io/nvidia/tensorrt:26.07-py3`
- TensorRT 11.1.0.106, CUDA 13.3, `torch==2.12.0+cu130` (matches
  `Dockerfile.dev.x86-gpu`, the community GPU CI environment)
- Model revisions: `openbmb/MiniCPM5-2B` (main), `TinyLlama/TinyLlama-1.1B-Chat-v1.0`
  (main) — both unpinned/latest at time of testing

### Not Run / Remaining Gaps

- Not validated against a real Llama 3.1+ checkpoint directly (gated on HF,
  requires license acceptance). Validated the mechanism generically (full
  id-list preservation, confirmed against a public mirror's `config.json`/
  `generation_config.json` shape for Llama-3.1-8B-Instruct) and validated
  end-to-end against MiniCPM5-2B's real multi-id config instead.
- Other existing llama e2e manifests (`falcon3-1b`, `minitron-4b-*`,
  `nemotron-nano-4b`) were not re-run against real checkpoints in this
  session; their code path is unchanged in kind (just array-wrapped), so
  risk is believed low but this is unverified.
- Tensor-parallel (`tensor_parallel_size > 1`) and dual-profile/dynamic-KV
  engine layouts were not separately exercised for the multi-id case (only
  the default "split" layout was tested); `eos_token_id` parsing happens once
  in `parse_runtime_config` before any layout-specific logic, so this is
  believed layout-independent but not explicitly run.

## Contributor Self-Review

Additionally, two independent rounds of focused review (correctness, test
coverage, contribution readiness) were performed; all findings from the
first round were addressed and re-verified in the second.

## Notes For Future Readers

- **Breaking change**: `runtime.json`'s `eos_token_id` field changes from
  sometimes-scalar to always-array. A bundle built with the pre-change llama
  plugin cannot be loaded by the post-change runtime, and vice versa (strict
  field parsing rejects the wrong shape). No migration path or version field
  is added: llama bundles are build-and-consume artifacts produced and
  immediately consumed by matching family code, not persisted or distributed
  release artifacts, and the repository has no existing convention for
  versioning a family section's internal field shapes. Anyone holding an
  existing llama bundle must rebuild it against this commit.
- **Out of scope, tracked separately**: `bos_token_id` (and `pad_token_id`)
  have the same missing-vs-zero bug the pre-fix `eos_token_id` had
  (`d.get(field, -1) or -1` maps a legitimate `0` to `-1`). Confirmed against
  MiniCPM5-2B's real config (`bos_token_id: 0`). Left untouched here to keep
  this PR single-purpose.
- **Found, NOT fixed here**: while regression-testing the ordinary
  single-EOS case, found that the existing `tinyllama-1.1b.json` e2e case
  fails against unmodified `upstream/main` — the native engine's first decode
  step predicts EOS immediately for a simple prompt, producing empty output,
  versus a correct HF reference ("Capital: Paris"). Reproduced identically
  against the unmodified upstream `config.py`/`model.py`/`plugin.cpp`,
  confirming this pre-dates this branch and is unrelated to `eos_token_id`
  handling (this PR's diff never touches chat-template detection/rendering
  code). Suspected root cause: `chat_templates.cpp`'s format-detection
  heuristic substring-matches on `<|user|>`/`<|assistant|>` tags and
  misclassifies TinyLlama's template as "Phi" format, injecting the wrong
  turn-separator token. Filing as a separate issue.
- Suggested review order: `config.py` → `model.py` → `plugin.cpp` → tests.

### Risk level

- [ ] Low
- [x] Medium
- [ ] High

Risk rationale: the bundle wire-format change is breaking but confined to an
internal build-and-consume artifact, not a persisted/distributed one; the
stop-token matching behavior change affects every llama-family model (not
just multi-EOS ones, since the field is now always array-wrapped), validated
broadly (16 unit/GPU tests + 1 real e2e correctness-oracle run) but not
against every existing llama manifest.


yifeif-nv:
Hi @roma5087 thanks for contributing. It looks like the community CI is failing. Please fix the community CI first and I can help you to trigger the internal CI next!

roma5087:
Closing this in favor of #1288, which landed while I was working through this — same bug (openbmb/MiniCPM5-2B's list-valued eos_token_id). It's additive (eos_token_ids only written when there's more than one id, so single-EOS bundles are untouched) rather than an unconditional breaking wire-format change like this PR's approach, and it adds vocab-bounds validation and explicit boolean rejection that this PR didn't have. Followed qwen3_8/model.py's existing precedent for the normalize pattern too.

PR #1280, opened 2026-09-12
build: Support x86_64 when using `docker compose`

zeroepoch:
You can use a command such as the following,

```docker compose build --build-arg ARCH=$(uname -m)```

to build the CI container for x86_64 (or aarch64).

- The default CPU architecture remains `aarch64` so CI doesn't break.
- Prefix `mpirun` with `HWLOC_COMPONENTS=-gl` to avoid a hang when graphics
  are present on the GPU.

## Background

The current `docker compose` workflow doesn't support anything but `aarch64` due to hardcoding some paths. Make the CPU arch a build argument to support `x86_64` as well.

## Exit Criteria

`docker compose build --build-arg ARCH=$(uname -m)` finishes successfully on a Linux x86 host (Fedora 44 + rootless podman) with an RTX 5080 GPU.

## Implementation

N/A: Just a build improvement.

### Change categories

- [ ] Public API
- [ ] ABI
- [ ] Dependencies
- [ ] Documentation only
- [x] CI or developer tooling

### Commands and Results

`docker compose build --build-arg ARCH=$(uname -m)`

- AMD Ryzen 9 5900X
- NVIDIA GeForce RTX 5080
- Fedora 44
- Podman

### Not Run / Remaining Gaps

N/A: Just a build improvement.

## Contributor Self-Review

## Notes For Future Readers

N/A: Just a build improvement.

### Risk level

This is a change only to the CI/developer Docker build, so it shouldn't have any product impact. The default value for the variable being added is the same as before so it should be a transparent change if the user doesn't exercise the variable.

- [x] Low
- [ ] Medium
- [ ] High

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@
 
 FROM nvidia/cuda:13.3.0-devel-ubuntu24.04@sha256:ef2203909e80b8b976cfc672f7e2ae2b00bc0e25c404ee86d89e10a3802f1c52
 
+ARG ARCH=aarch64
+
 ENV DEBIAN_FRONTEND=noninteractive
 ENV TORCH_CUDA_ARCH_LIST=10.0
 ENV OMPI_ALLOW_RUN_AS_ROOT=1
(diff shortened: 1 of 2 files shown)


yifeif-nv:
Hey Eric, please add a short explanation under the risk level. 

There is also another failure, which is a gap in our CI: we need to add the docker-compose.yml to shared files in tools/test_impact.py and we should be good here

zeroepoch:
> Hey Eric, please add a short explanation under the risk level.

Added

> There is also another failure, which is a gap in our CI: we need to add the docker-compose.yml to shared files in tools/test_impact.py and we should be good here

I'll take a look at this and update the PR.

PR #1131, opened 2026-09-02
feat(qwen): add s1-mini-fp16 manifest and validation config

AbishekCoder1:
## Background

Add S1-mini (by Superwhisper), an ASR text-normalizer fine-tune of Qwen3-0.6B. Originally added as a manifest inside the `qwen` family; per maintainer review (@yifeif-nv), restructured into its own family, `families/s1_mini/`, since S1-mini is a distinct ASR-based model despite sharing the Qwen3 architecture.

## Exit Criteria

- `families/s1_mini/` exists as a complete, independent family per repo architecture rules
- `s1-mini-fp16` manifest and threshold sidecar are correctly schemed
- `tools/tests/test_architecture.py` passes in full for the new family
- Release-performance suite coverage is honest about current evidence

## Implementation

- Created `families/s1_mini/` by porting `families/qwen/`'s Python build graph and native C++ runtime (model, config, checkpoint mapping, graph builders, native KV/sampler/tokenizer/CMake, and cpp unit tests), since S1-mini uses the same Qwen3 architecture. Renamed family identifiers throughout (`family="s1_mini"`, CMake targets, include paths); left internal Qwen3-specific algorithm names (`QwenSamplingParams`, `qwen_expand_layer_name`, etc.) untouched, since they're implementation details of the shared architecture, not the family boundary.
- `families/s1_mini/tests/manifests/s1-mini-fp16.json`: corrected to current schema (`task`, `max_sequence_length`, `tensor_parallel_size`; no forbidden central-orchestration keys); sends the model's documented system+user prompt split; added `premerge: true`.
- `families/s1_mini/tests/thresholds/s1-mini-fp16-e2e.json`: removed unused logit-oracle overrides (testcase isn't logit-registered); tightened `normalized_text_edit_distance` from an accept-anything `1.0` to `0.35`; added `expected_answers: ["Thursday"]` as a real domain assertion on the self-correction behavior.
- `apps/benchmark/performance/release.yaml`: excluded `s1-mini-fp16` with an honest reason (no workload/receipt exists yet), rather than the earlier wording that overclaimed parity qualification.

### Change categories

- [x] Model or runtime behavior
- [ ] Public API
- [ ] ABI
- [x] Bundle or artifact format
- [ ] Dependencies
- [ ] Documentation only
- [x] CI or developer tooling

### Commands and Results
45/45 passed.
1 passed.

### Hardware, Environment, and Revisions

[...]

PR #1283, opened 2026-09-12
fix(flux): check the return value of ensure_buf's cudaMalloc

Moviw:
## Background

`flux_gpu_matmul`'s three module-scope scratch buffers (`g_dA`, `g_dB`, `g_dC`) are grown lazily by `ensure_buf`, which frees the too-small buffer and calls `cudaMalloc` without checking the returned `cudaError_t`. Same shape as the bug fixed in `bark` by #1201, reported for six sibling families (including `flux`) by #1221. On an allocation failure the pointer stays null while `bytes` is already updated to the requested size, so the next call sees the buffer as correctly sized and the first real use (`cudaMemcpyAsync`/`cublasSgemm` into a null pointer) fails opaquely, well past the point where the allocation failure was known.

## Exit Criteria

`ensure_buf`'s `cudaMalloc` call is checked; a failed grow throws instead of silently marking the buffer as sized. Success path (buffer reuse when already large enough, growth when not) is unchanged.

## Implementation

**Model/component:** `flux` family only (`families/flux/runtime/`), matching this repo's one-family-per-PR convention — the other five families from #1221 are separate PRs (`bart`/`whisper` merged as #1222/#1223, `m2m_100`/`marian`/`t5` open as #1232/#1233/#1234).

Introduces `families/flux/runtime/device_buffer.h` with a `DeviceBuffer` RAII wrapper (same shape as the one already merged for `bart`/`whisper` and open for the other three families) and a `GrowableBuffer` + `ensure_buf` pair specific to flux's shape: unlike the per-request buffers the sibling families own, flux's three scratch buffers are process-lifetime globals reused and grown across calls. `ensure_buf` now frees the stale buffer, checks `cudaMalloc`'s result, throws `std::runtime_error` on failure, and — critically for the reused-buffer shape — leaves `bytes` unchanged on failure so the next call retries the grow instead of treating a missing buffer as already sized.

`flux_gpu_matmul_shutdown()`'s manual `free_buf` lambda is replaced by move-assigning a fresh `GrowableBuffer` to each global, which is now sufficient since `DeviceBuffer`'s destructor/move-assignment already release the device allocation.

No public API, ABI, or bundle/artifact change — this is an internal, family-local error-handling path.

### Change categories

### Commands and Results

The real `trtmc_model_flux` target needs the TensorRT SDK, which is not installed in this environment (see Not Run below), so `ensure_buf`/`DeviceBuffer` were validated the same way #1222/#1223 validate the analogous fix: a CPU-only, no-GPU-required test built and run directly (the repo's own CMake/ctest wiring for it is included but not exercised here — see Not Run):

```
$ g++ -std=c++17 -Wall -Wextra -Wpedantic -I. -Icore/runtime/include \
    families/flux/tests/cpp/test_flux_device_buffer_alloc.cpp \
    -o /tmp/test_flux_device_buffer_alloc
$ /tmp/test_flux_device_buffer_alloc
flux device buffer allocation-failure checks passed
```

The test covers: (1) a no-op grow when the request already fits does not call the allocator; (2) a failed grow throws, leaves the pointer null instead of stale, leaves `bytes` unchanged, and releases the old buffer instead of leaking it; (3) retrying after a transient failure succeeds normally.

Mutation check — confirmed the test actually catches the original bug: temporarily reverted `ensure_buf` to the old unchecked shape (`gb.buf.allocate(need); gb.bytes = need;`, no error check) and reran:

```
FAIL: a failed grow should throw
FAIL: a failed grow must not update the tracked size
2 check(s) failed
```

Restored the fix and reran; passes cleanly again (output above).

`gpu_matmul.cpp` itself was reviewed by hand and by `grep` for any remaining reference to the old `DevBuf`/raw-`.ptr` shape (none found) but was not compiled — see Not Run.

Repo head: this branch off `upstream/main` at `474c50e7` (`feat(qwen): add s1-mini-fp16 manifest and validation config (#1131)`). Host: g++ 11.4.0, CUDA 11.5 (`nvcc`)/CUDA 13.0 driver 580.173.02, RTX 3090 present but TensorRT SDK not installed on this box (would come from the `Dockerfile.community-cpu` image's `libnvinfer-dev` packages, not present locally).

### Not Run / Remaining Gaps

- Did not build `trtmc_model_flux` or run it through the repo's own `cmake --build`/`ctest`: this environment has no TensorRT SDK installed (`find_path(TRTMC_TRT_INCLUDE_DIR ...)` fails), so the full project cannot configure here. The CI "Community CPU" check should exercise this properly; flagging so a reviewer knows the CMake/ctest wiring in this PR (mirroring #1222/#1223's) is untested by me locally, only the test binary's direct compilation.
- No GPU/inference run of `flux_gpu_matmul_bias` itself (needs the built shared library + real cuBLAS linkage): the change to that function is a mechanical pointer-type change (`.ptr` → `static_cast<float*>(...get())`), not a logic change, so I relied on the CPU-level `ensure_buf`/`DeviceBuffer` test plus manual review rather than an end-to-end run.
- No real Flux checkpoint/engine on hand to exercise the preprocessor path this buffer feeds.

## Contributor Self-Review

Self-reviewed against `plugins/trtmc-agent-skills/skills/review-trtmc-pr`'s manual probes: family-isolated (only `families/flux/**` touched, no shared/core file), no new public API/ABI/shared abstraction, no cross-family import, `grep`-verified no stale references to the old shape remain. One behavior note volunteered under Notes below (harmless double-release-on-process-exit) rather than left for a reviewer to find.

## Notes For Future Readers

Minor, harmless behavior note: previously `DevBuf` was a POD with no destructor, so only the explicit `flux_gpu_matmul_shutdown()` freed the buffers. `GrowableBuffer` now has a real destructor (via `DeviceBuffer`), so process exit will also run `cudaFree` on whatever `flux_gpu_matmul_shutdown()` already reset to null — `cudaFree(nullptr)` is a documented no-op, so this is inert, just flagging the shape change for whoever reads this next.

Companion fixes for the other five families reported by #1221: `bart`/`whisper` merged (#1222/#1223), `m2m_100`/`marian`/`t5` open (#1232/#1233/#1234).

### Risk level

- [x] Low

Error-handling-only change on an existing, already-private allocation path; no public API, ABI, or success-path behavior change. Verified by direct compilation + a mutation check that the new check actually fires; not verified against the real GPU/TensorRT build in this environment.


yifeif-nv:
Hey @Moviw thanks for contributing. Looks like the community CI has failed on the source quality check. Please help to fix that and I'll trigger the internal CI for you. General direction of this PR looks good

Moviw:
Thanks for the quick review! Fixed both:

- **Source quality**: `clang-format` wanted a blank line separating `gpu_matmul.h` (the file's own header) from the other quoted include — fixed and verified clean locally (`clang-format --dry-run --Werror` on all three touched files).
- **DCO**: the commit was missing `Signed-off-by` — amended and force-pushed with `--signoff`.

Should be green now, or ready for the internal CI trigger whenever you get a chance.

PR #1326, opened 2026-09-17
feat(qwen3_8): wire real NVFP4 + FP8 GEMM via add_dynamic_quantize

zhenshanx-nv:
## Background

`families/qwen3_8` (RadixArk/Qwen3.8-27B-NVFP4-style ModelOpt MIXED_PRECISION
checkpoints) shipped real packed NVFP4 (W4A4) weights for MLP/lm_head and FP8
(W8A8) weights for attention/DeltaNet projections, but the engine builder
unconditionally dequantized everything to FP16/FP32 at load time and
hard-rejected any quantized build request (`NotImplementedError` in
`build_engine()`). NV Blackwell (SM100) hardware natively supports W4A4/W8A8
tensor-core GEMM, so the checkpoint's own quantized weights should drive real
FP4/FP8 GEMM kernels instead of being fully dequantized and run as plain FP16
matmuls.

TensorRT's public `add_quantize` + `block_shape` API is unconditionally
rejected for FP4 output (`shape.cpp:3350`, "Blockwise quantization
requires output type to be int8 or fp8e4m3", confirmed on TensorRT
11.1.0.106 and 11.3.0.99). `IDynamicQuantizeLayer`
(`network.add_dynamic_quantize`) is the one public layer type with a real
fused FP4 tensor-core kernel and is not blocked by that check.

## Exit Criteria

- `qwen3_8` builds a real engine for `RadixArk/Qwen3.8-27B-NVFP4` with
  `--quantization nvfp4` instead of raising `NotImplementedError`.
- Checkpoint NVFP4/FP8 weights are fed to TensorRT as-is (bit-exact packed
  bytes + their own calibrated scales), not dequantized-then-requantized.
- The engine inspector shows the quantized constants really are FP4E2M1 /
  FP8 typed, and real fused Blackwell tensor-core kernels are selected for
  those layers (not a dequant-to-FP16 fallback).
- End-to-end generation with the real checkpoint and its own tokenizer/chat
  template produces coherent, correct output.
- Non-goal: DeltaNet's `a_proj`/`b_proj` stay unquantized (matches checkpoint
  layout); this PR does not add a new quantization scheme, only wires the
  checkpoint's existing one through to real GEMM kernels.

## Implementation

New `families/qwen3_8/quantization.py`:
- `calibrate_qwen3_8_nvfp4(model_dir, config, graph_ops)` reads the
  checkpoint's own packed NVFP4 bytes (MLP gate/up/down, lm_head) and FP8
  bytes (DeltaNet in_proj_qkv/in_proj_z/out_proj, attention q/k/v/o) plus
  their real calibrated `weight_scale`/`weight_scale_2`/`input_scale`
  tensors directly, returning a `Qwen38QuantContext`.
- `_NVFP4Format.wrap_matmul`: weight side uses only `add_dequantize` (two
  levels: block scale via `weight_scale_2`, then per-16-element
  `weight_scale`) on the checkpoint's native packed FP4 constant --
  bit-exact reuse, no requantization. Activation side uses
  `add_dynamic_quantize` (block size 16) since that is the only public API
  path to a real fused FP4 GEMM prologue.
- `_FP8Format.wrap_matmul`: standard `add_quantize`/`add_dequantize`
  per-tensor pattern (no blockwise restriction applies to FP8), matching
  `families/qwen`'s already-proven FP8 approach.
- Weight constants are fed in native `[out_features, in_features]`
  checkpoint layout with `MatrixOperation.TRANSPOSE` on the matmul, to avoid
  unpacking/transposing/repacking sub-byte packed FP4 data.

`families/qwen3_8/engine_builder.py`:
- Removes the `quant_ctx` `NotImplementedError` guard in `build_engine()`.
- Threads `quant_ctx` through DeltaNet (`in_proj_qkv`/`in_proj_z`/`out_proj`)
  and attention (`q`/`gate`/`k`/`v`/`o`) matmuls via the existing
  `graph_blocks.make_matmul_fn` (already quant_ctx-aware, shared with the
  MLP/lm_head path).
- Includes a required precision-threading fix: `_transpose_2d`'s
  `precision` parameter was never passed through
  `load_weights()`/`_load_*_weights()`, so every weight was stored as FP32
  regardless of `--precision`, which OOMs a full 27B build. This exact fix
  also lives isolated on `zhenshanx/qwen3_8-precision-threading-fix` for
  landing as its own PR; drop this hunk on rebase once that merges.
- Sets `ProfilingVerbosity.DETAILED` for engine-inspector tactic/constant
  visibility (diagnostics only, no behavior change).

`families/qwen3_8/model.py`:
- Accepts `quantization="nvfp4"`, builds `quant_ctx` via
  `calibrate_qwen3_8_nvfp4()`, threads `precision` into `load_weights()`.

No public API, ABI, or bundle-format changes: quantized engines are only
produced when `--quantization nvfp4` is explicitly requested; the
unquantized build path is unchanged.

### Change categories

### Commands and Results

Python-path build + engine inspection (`nvfp4_venv`, TensorRT 11.1.0.106):

```
python3 test_full_build.py   # drives Qwen38Model.build() against the real checkpoint
# -> engine.plan written, 20,165,899,580 bytes (down from ~53.8GB naive-quantize,
#    and ~54GB unquantized FP16 baseline)
```

Engine inspector (`IEngineInspector`, `LayerInformationFormat.JSON`,
`ProfilingVerbosity.DETAILED`) constant-type census: 193 FP4E2M1-typed and
546 FP8-typed constants, matching every quantized `weight_name` registered
by `calibrate_qwen3_8_nvfp4`. Tactic names show real Blackwell tensor-core
kernels (`tensorop*`/`cga*`/`sm*`), automatic fusion of gate+up MLP
projections into `dual_gemm_fused`, and `CastMulCast`->`Fc` fused
FP8-quantize-prologue+GEMM kernels -- not a dequant-to-FP16 fallback.

Hand-rolled single-step decode driver (matches the C++ runtime's exact
attention-mask/state/position semantics) with the real
`RadixArk/Qwen3.8-27B-NVFP4` tokenizer and chat template:

```
Prompt: "What is the capital of France? Answer in one word."
Output: "Paris<|im_end|>"
```

Native `trtmc` C++ runtime (source-built against
`nvcr.io/nvidia/tensorrt:26.07-py3`, targets `trtmc`, `trtmc_backend_trt`,
`trtmc_model_qwen3_8`, `trtmc_dataset_benchmark`), full bundle built via
`python -m tensorrt_model_connect build ... --quantization nvfp4` (bundle
size 20,188,813,008 bytes, matching the Python-path engine size) and run
through `trtmc_dataset_benchmark`:

```
[trtmc-perf] Prefill: 14 tokens, 149.7 ms (93.5 tok/s)
[trtmc-perf] Decode:  60 steps, 503.4 ms (119.2 tok/s, 8.39 ms/tok)
```
Output text: coherent, on-topic completion for a real generation prompt
("Explain what a large language model is, in about five sentences.").

Build-time resource cost (Python path, `test_full_build_instrumented.py`,
stage-by-stage RSS sampling):
- `calibrate_qwen3_8_nvfp4`: +2GB CPU RSS (real calibration data only).
- `load_weights`: +55GB CPU RSS (dominated by the known
  precision-threading bug fixed here, and by the tracked-but-not-yet-fixed
  redundant double-dequantization of weights `quant_ctx` already owns --
  see Notes below).
- `build_engine`: +43GB CPU RSS, ~19.2GB peak GPU memory (genuine TensorRT
  builder overhead).

- GPU: B100 / GB100-class, SM100 (compute_cap 10.0), 183GB VRAM.
- TensorRT: 11.1.0.106 (both the pip wheel used for the Python-path
  validation and the NGC `tensorrt:26.07-py3` container used for the
  native-runtime build/benchmark).
- Checkpoint: `RadixArk/Qwen3.8-27B-NVFP4` (real HF checkpoint, not a
  synthetic/tiny fixture).
- Repository head at validation time: `d7b3472a` (this branch, rebased onto
  `main`).

### Not Run / Remaining Gaps

- No automated `test_e2e.py`/`test_engine_builder.py`/
  `test_checkpoint_mapper.py` coverage added for the new NVFP4/FP8 path in
  this PR -- all validation above was manual, on real hardware and the real
  checkpoint. Adding an automated E2E/unit test is a reasonable fast-follow.
- CUDA graph capture was not exercised for the native-runtime benchmark
  numbers above (matches the comparison baseline used).
- Multi-batch / concurrent-request throughput was not measured (only
  batch-size-1 decode).

## Contributor Self-Review

## Notes For Future Readers

- The precision-threading fix bundled into this PR (see Implementation) is
  also tracked as its own isolated fix on
  `zhenshanx/qwen3_8-precision-threading-fix` for a follow-up PR; once that
  lands and this branch is rebased, drop the duplicated hunk here.
- Known, tracked-but-not-fixed-here inefficiency: `load_weights()` still
  loads and fully dequantizes checkpoint tensors for every
  `weight_name` that `quant_ctx` already owns, even though
  `maybe_quantized_matmul()` never uses that dequantized copy for those
  weights. This wastes CPU time/memory during build (roughly half of the
  `load_weights` RSS growth measured above) but is not a correctness issue.
  Fixing it requires passing `quant_ctx` into `load_weights()` and skipping
  `_load_tensor` for names it owns, which changes the current call order in
  `model.py` -- left for a follow-up perf-only PR.
- Suggested review order: `quantization.py` (new mechanism) first, then the
  small `engine_builder.py`/`model.py` threading changes.

### Risk level

- [x] Medium

Risk rationale: touches production inference behavior for a real, shipping
quantized checkpoint family, but is strictly additive/opt-in (only active
under `--quantization nvfp4`) and the existing unquantized build/run path is
unchanged and untouched by this diff.


zhenshanx-nv:
Update: `nvidia/Qwen3.8-27B-NVFP4` (NVIDIA's own official release of this checkpoint) is now the checkpoint tested against and recommended, replacing `RadixArk/Qwen3.8-27B-NVFP4` referenced above. Both are the same ModelOpt MIXED_PRECISION export -- verified byte-identical in structure and behavior (weight loading, build time, engine size, runtime, accuracy) -- so no code change was needed; `calibrate_qwen3_8_nvfp4()` never hardcoded either checkpoint name. The code comments/docstrings that still named RadixArk's checkpoint were updated in #1389.

PR #1327, opened 2026-09-17
feat(gemma): qualify gemma-3-270m and refuse fp16 for Gemma 3

zhenshanx-nv:
## Background

#1304 added the Gemma 3 text path and qualified the 1B and 4B. Google publishes three more
text-capable widths; this qualifies two of them on that same path, with no code change to the
builder, and closes off a precision that is not safe for the generation.

`gemma-3-270m` is 268M — the smallest useful model this repository builds. `gemma-3-12b` is
structurally identical to the already-qualified 4B: same `Gemma3ForConditionalGeneration`,
nested `text_config`, `sliding_window` 1024, pattern 6, `rope_local_base_freq` 10000,
`rope_theta` 1e6, linear rope scaling factor 8, `query_pre_attn_scalar` 256 — only wider
(48 layers, hidden 3840, 16 heads, 8 KV heads).

## Exit Criteria

- Both widths match the Hugging Face reference, on the existing code path.
- A precision that is wrong for this generation cannot be selected.
- Gemma and Gemma 2 are unaffected.

## Implementation

Manifests only for the two widths. The one code change is the fp16 refusal.

### Change categories

### Commands and Results

Both built at bf16 and driven through the native CLI against the Hugging Face reference,
using the prompt rendering and metric of `families/gemma/tests/test_e2e.py`:

| width | engine | reference | NED |
| --- | --- | --- | --- |
| gemma-3-270m | `[50429, 107, 106]` `Paris` | identical | **0.0000** |
| gemma-3-12b | `[50429, 106]` `Paris` | identical | **0.0000** |

**Why fp16 is refused.** Largest absolute value leaving a decoder layer, reference in fp32,
against the fp16 maximum of 65,504:

| width | peak | layers over | fp16 result |
| --- | --- | --- | --- |
| gemma-2-2b | 4,059 | 0 of 26 | works, 16x headroom |
| gemma-3-270m | 102,956 | 11 of 18 | token 0 repeatedly |
| gemma-3-1b | 61,040 | 0 of 26 | works, by 7% |
| gemma-3-4b | 298,680 | 29 of 34 | token 0 repeatedly |

The 1B fits on one prompt by 7%, which is luck rather than headroom, so the refusal covers the
generation rather than individual widths. Gemma 2 keeps fp16 on its measured margin.

Other suites:

- `families/gemma/tests/` — 35 passed, 2 of them new
- `apps/benchmark/trtmc_benchmark/tests` — 1107 passed
- `ruff check` clean

- NVIDIA H100 80GB HBM3, x86 dev container
- Manifests name the official `google/gemma-3-270m-it` and `google/gemma-3-12b-it`. Those repos
  are gated and this container has no HF token, so local verification used the `unsloth`
  mirrors of the same weights; CI fetches the official ones, as it already does for
  `gemma-2-2b-it` and the widths in #1304.
- Rebased on `29facea2`

### Not Run / Remaining Gaps

- **`gemma-3-27b` is deliberately not in this PR.** It is the last text width and needs a
  large-memory host to build — the comparable Qwen3.6-27B build peaked at 291 GB — so it is
  being qualified separately rather than shipped unverified here.
- Gemma 3n and Gemma 4 remain refused by the model-type gate.
- Both widths are in `excluded_profiles`, not benchmarked: the release-performance receipt was
  collected only for `gemma-2-2b`, which exercises the same builder and runtime path.

## Contributor Self-Review

## Notes For Future Readers

The fp16 numbers are the part worth keeping. Gemma 3 was qualified at bf16 in #1304 after fp16
produced a single repeated token, but the reason was not established at the time. Measuring the
activations settles it: this is range, not a graph defect, and the 1B "working" at fp16 is an
accident of one prompt landing 7% inside the limit. Refusing per width would have shipped that
accident as a supported configuration.

Third-party provenance: Gemma is published by Google; the checkpoints are the
`google/gemma-3-270m-it` and `google/gemma-3-12b-it` Hugging Face repositories.

### Risk level

- [x] Low

Two manifests on an existing code path, plus a refusal that only narrows what can be built.
No checkpoint that builds correctly today is affected.


zhenshanx-nv:
Dropped `gemma-3-12b` from this PR.

The internal premerge run on `5d5680c8` failed only on that case:

```
E  AssertionError: assert 0.8387096774193549 <= 0.15
E   +  where 0.8387096774193549 = _normalized_edit_distance('The capital of France is Paris.', 'Paris')
families/gemma/tests/test_e2e.py:480
```

The HF reference produced `Paris`; the engine produced a different continuation. My local
qualification of the same manifest passed exactly (`engine ids=[50429, 106] text='Paris'`,
`reference ids=[50429, 106] text='Paris'`, NED=0.0000), but it ran against the
`unsloth/gemma-3-12b-it` mirror because `google/gemma-3-12b-it` is gated. Divergence between
the official and mirrored `config.json` is the leading explanation — the same class of issue
produced three earlier rounds of failures on #1304 — but I have not been able to confirm it,
so the manifest is withheld rather than retried blind.

`gemma-3-270m` and the fp16 refusal are unchanged and independent of this.

PR #1348, opened 2026-09-18
DO NOT MERGE: diagnostic for the gemma-3-12b CI divergence

zhenshanx-nv:
## Background

`gemma-3-12b` was dropped from #1327 because the internal premerge run failed only on that
case: the engine produced `The capital of France is Paris.` where the HF reference produced
`Paris`. The argument order in `families/gemma/tests/test_e2e.py:452-480` confirms the first
string is the engine output, so it is the engine that is wrong, not the reference.

The case passes locally, exactly, against the `unsloth/gemma-3-12b-it` mirror, which is what
local runs must use because `google/gemma-3-12b-it` is gated and this environment has no HF
token. Four hypotheses have been eliminated locally:

1. Config divergence between the mirror and the official repo. Two independent mirrors,
   including a finetune derived from the official repo, match unsloth on every Gemma 3 field.
2. A near-tie at the first generated token. The fp32 reference gives a top1/top2 margin of
   23.69 logits with `p(top1) = 1.0000`; `The` is not in the top 8.
3. GPU architecture. Rebuilt on H100/sm90 as well as the original A100/sm80; both give `Paris`.
4. The chat template not being applied. Feeding the raw prompt yields
   `\nParis\nWhat is the highest mountain`, which is not the CI text either, and
   `families/gemma/runtime/pipeline.cpp:91-96` raises rather than silently skipping when the
   template is absent.

The only remaining difference is the checkpoint itself, which I cannot read from here.

## Exit Criteria

The internal run logs `[GEMMA-DIAG]` lines showing what the builder resolved from the official
checkpoint, so those values can be compared against the mirror. This branch is then deleted and
any real fix goes on its own pull request. Non-goal: merging any part of this.

## Implementation

Restores the `gemma-3-12b` manifest and its `excluded_profiles` entry, and adds temporary
build-time prints in `families/gemma/model.py` covering model type, layer and head geometry,
sliding window and pattern, both rope bases, `query_pre_attn_scalar`, and the full raw key list.

### Change categories

- [x] CI or developer tooling

### Commands and Results

Local, against `unsloth/gemma-3-12b-it`, repeated on two architectures:

```
engine    ids=[50429, 106] text='Paris'
reference ids=[50429, 106] text='Paris'
NED=0.0000 (threshold 0.15)  PASS
```

First-token margin from the fp32 reference:

```
50429  84.3816 'Paris'
142253 60.6887 '巴黎'
top1-top2 margin = 23.6929   p(top1) = 1.0000
```

Engine with the chat template disabled, to test hypothesis 4:

```
text="\nParis\nWhat is the highest mountain"
```

Branched from `upstream/main` at e48814b6. Checkpoint `unsloth/gemma-3-12b-it` (the gated
`google/gemma-3-12b-it` is unavailable here). H100 80GB HBM3 and A100 80GB PCIe, TensorRT
backend, bf16 build with an fp32 reference, CUDA arch sm90 and sm80.

### Not Run / Remaining Gaps

Everything against the official gated checkpoint, which is the entire point of this branch. No
new tests are added and no fix is proposed yet.

## Contributor Self-Review

## Notes For Future Readers

The mirror-versus-gated-checkpoint gap is a recurring hazard for Gemma work in this repo: three
earlier rounds of failures on #1304 came from fields Google omits that mirrors fill in. Note
that the mirror comparison above argues against that explanation this time, so if the diagnostic
comes back identical, suspicion moves to the weights rather than the config.

### Risk level

- [x] Low

Draft, diagnostic only, and must not merge. The prints are removed with the branch.


zhenshanx-nv:
Diagnostic complete. The [GEMMA-DIAG] output showed google/gemma-3-12b-it states only geometry and relies on Gemma3TextConfig for the rest, which left `rope_theta` at the parser's 1e4 fallback and collapsed the attention schedule to all-global. Fixed in #1362, which also reproduces the failure locally by reconstructing this config shape over mirror weights. Closing; not for merge.

